PR #30, opened 2018-01-18
Support setting environment variables. Closes thenativeweb/org#516.

grundhoeferj:


lib/application/getConfiguration.js
```diff
@@ -55,6 +56,18 @@ const getConfiguration = async function (options) {
...
   }
 
+  Object.keys(configuration.environments).forEach(name => {
+    const currentEnvironment = configuration.environments[name];
+
+    if (!currentEnvironment.environmentVariables) {
+      return;
+    }
+
+    const { environmentVariables } = currentEnvironment;
```
goloroden: Wouldn't it make sense to move this to the line *before* the `if`?

lib/application/transformEnvironmentVariables.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const snakeCase = require('lodash/snakeCase');
+
+const transformEnvironmentVariables =
```
goloroden: There something with the indentation is wrong – or, I'd suggest to move the line with `function` to the line of the assignment.

lib/application/getConfiguration.js
grundhoeferj: Sure! 😊

lib/application/transformEnvironmentVariables.js
grundhoeferj: Hups... 🤔 it is fixed now!

grundhoeferj:
I fixed the issues 😊

PR #32, opened 2018-02-05
Add force flag to init command.

grundhoeferj:


test/stories/setupTests.js
```diff
@@ -112,5 +113,39 @@ const isolatedAsync = promisify(isolated);
...
+
+      const { code, stderr, stdout } = await wolkenkit('init', { force: true }, { cwd: directory });
+
+      assert.that(code).is.equalTo(0);
+      assert.that(stdout).is.equalTo(`  Initializing a new application...\n  Cloning ${template}...\n✓ Initialized a new application.\n`);
+      assert.that(stderr).is.equalTo('');
+
+      const directoryList = await getDirectoryList(directory);
+
+      assert.that(directoryList).is.containingAllOf([ 'client', 'server', 'package.json', '.git' ]);
```
goloroden: Maybe we could add some additional check here, to ensure that the `.git` directory and its contents have not been changed.

lib/wolkenkit/commands/init/index.js
```diff
@@ -32,35 +37,72 @@ const init = async function (options, progress = noop) {
...
 
-    throw new errors.DirectoryNotEmpty();
-  }
+      throw ex;
+    }
+  };
 
-  if (!await shell.which('git')) {
-    progress({ message: 'git is not installed.', type: 'info' });
+  if (force) {
```
goloroden: Now that I reviewed this code once again, I thought that it may be a good idea to move this large `if` block to its own function. What do you think?

lib/wolkenkit/commands/init/index.js
```diff
@@ -1,15 +1,17 @@
...
+const isolated = require('isolated'),
+      promisify = require('util.promisify'),
+      recursiveReaddir = require('recursive-readdir');
 
 const errors = require('../../../errors'),
       noop = require('../../../noop'),
       shell = require('../../../shell');
 
-const readdir = promisify(fs.readdir);
+const isolatedAsync = promisify(isolated),
```
goloroden: Shouldn't we change this to

```javascript
const isolated = promisify(isolatedCallback);
```

?

I.e., change the original function name to be suffixed with `Callback` (or `Async`), and then use the promisified one as it was originally intended? I mean, when promisifying `fs.readdir` we don't call it `readdirAsync` either 😉

grundhoeferj: Sure..! ;-)

lib/wolkenkit/commands/init/index.js
grundhoeferj: That is a good idea! 😊

test/stories/setupTests.js
grundhoeferj: Done! 😊

grundhoeferj:
Here is a new version. Ask for review, please 😊

PR #44, opened 2018-08-10
Limit number of restarts. Validate container logs for better error messages.

grundhoeferj:


src/wolkenkit/commands/shared/validateLogs.js
```diff
@@ -0,0 +1,102 @@
...
+            reject(ex);
+          }
+        });
+      } catch (ex) {
+        isStopped = true;
+
+        validate.emit('error', ex);
+      }
+
+      await sleep(250);
```
goloroden: Do you remember that line of code (somewhere) where we had a `setTimeout`, but couldn't remember what it was good for? This could become a similar place 😉 

Maybe it would be a good idea to add a comment here on *why* the waiting is necessary.

grundhoeferj:
@goloroden Done 😊

PR #48, opened 2018-09-13
Introduce wolkenkit-secrets.json.

grundhoeferj:


src/application/resolveSecrets.js
```diff
@@ -0,0 +1,63 @@
...
+    secrets = await file.readJson(path.join(directory, 'wolkenkit-secrets.json'));
+  } catch (ex) {
+    if (ex.code !== 'EFILENOTFOUND') {
+      throw ex;
+    }
+  }
+
+  const identifier = 'secret://';
+
+  const flattenConfiguration = flatten(configuration);
```
goloroden: This sounds as if `flattenConfiguration` was a function to flatten the configuration. What you mean, I think, is `flattenedConfiguration`, isn't it?

src/application/resolveSecrets.js
```diff
@@ -0,0 +1,63 @@
...
+
+  Object.keys(flattenConfiguration).forEach(key => {
```
goloroden: This could be replaced by a `for ... of` loop with `Object.entries`, which not only delivers the key, but at the same time also the value:

```javascript
for (const [ key, value ] of Object.entries(flattenConfiguration)) {
  // ...
}
```

src/application/resolveSecrets.js
```diff
@@ -0,0 +1,63 @@
...
+
+    if (!secrets) {
+      throw new errors.SecretFileNotFound('wolkenkit-secrets.json is missing.');
+    }
+
+    const secretSelector = value.replace(identifier, '');
+    const secret = get(secrets, secretSelector);
+
+    if (secret === undefined) {
+      throw new errors.SecretNotSpecified(`Could not find a secret named '${secretSelector}'.`);
```
goloroden: Shouldn't this be `SecretNotFound` instead of `SecretNotSpecified`?

grundhoeferj:
Things are fixed! 😊

PR #60, opened 2018-11-09
Introduce axios and ping root.

goloroden:
Hi @grundhoeferj 😊 

This morning, I was up early, and I removed `request` and `requestretry`, and instead added `axios` and `async-retry`.

Besides, I also changed that we no longer wait for the `/ping` route, but now just for root, as this is the endpoint you can also access in the browser.

Please do *NOT YET*  merge this, as we still need to run the tests and check backwards compatibility, but what I wanted you to ask was if you could just review it, especially the cases that are related to aufwind, if I missed anything.

Thanks 😊

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,9 +105,7 @@ $ npx roboter test-stories
 
 - Running the story tests getting an error called *Error loading extension section V3_ca*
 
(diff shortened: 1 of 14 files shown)


goloroden:
The tests all run now 😊 

So we are basically ready for review.

PR #136, opened 2019-02-28
Improve error output when a CLI command fails. Closes #135

### LINKED ISSUE #135: Improve error output when a CLI command fails
### What is this concept about?

Currently when a CLI command like `wolkenkit start` fails, there is no output about the actual error that caused the command to fail. This is very annoying and also makes failing commands harder to debug as you need to run them again.

I think we should switch this default behaviour. So when an error occurs, the details about this error should always be displayed, just like when running in `verbose` mode.

### What issues need to be clarified to complete this concept?

- [x] Figure out if this somehow may cause issues with `--verbose` mode
- [x] Enable error output for failing commands in the CLI

goloroden:
Hi @mattwagl 😊 

I have changed the error output so that the CLI now always shows the verbose information with the error details, no matter whether verbose mode was enabled or not. This should solve #135, and improve usability.

Can you please have a look at it, and if you are happy with it, squash and merge it?

I have already run all the tests (even the ones on AWS), and everything works so far 😊

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -891,9 +891,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "10.12.18",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-10.12.18.tgz",
-      "integrity": "sha512-fh+pAqt4xRzPfqA6eh3Z2y6fyZavRIumvjhaCL753+TVkGKGhpPeyrJG2JftD0T9q4GF00KjefsQ+PQNDdWQaQ==",
+      "version": "10.12.27",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-10.12.27.tgz",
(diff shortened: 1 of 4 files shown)


mattwagl:
Thanks for responding so quickly! :-) I tested it on my local machine by logging into Docker using my e-mail, and running `wolkenkit start` and it failed. I got the error message, which is nice in this case. :-)

One little notice: currently when installing wolkenkit, we get an npm warning npm due to a critical vulnerability in the `open` package which also seems to be pretty inactive. Fortunately there is a more active alternative called [`opn`](https://www.npmjs.com/package/opn) :-) I think it would be worth to switch since it gives a really bad impression when installing the CLI.

What do you think?

goloroden:
First, we have resolved the merge conflicts.
Second, we have replaced `open` by `opn`.

Thanks for the hint, and I'm glad you like it 😊 

If you want to, you can squash / merge it now.

PR #160, opened 2019-06-30
Lockstore implementations

damienbenon:
Brings implementations for:
* MySql
* MariaDb
* MongoDb
* Postgres
* Redis
* SqlServer.

All implementations pass Integration and Unit tests.
During development, the integration test `renewLock#'renews the lock.'` has been know to be a bit flaky with MariaDb/MySql implementations.

After some interesting reading about distributed locks and this [very interesting article by Redis](https://redis.io/topics/distlock), it may be possible that the current implementation is a bit loose and does not fit 100% of the final requirements of wolkenkit.

I would recommend to use preferably the `Redis` lockstore implementation as this is the closest of the recommended redlock algorithm (although working only on a single node). 

The `username` provided during the `initialize` of `RedisLockstore` is used as the `value` along with the hashed key to validate that the lock is indeed held by that lockstore instance.. Thus providing unique names can guarantee on different operations (I mainly have `renewLock`) that no other Lockstore instance is trying to hijack a current lock.


damienbenon:
BTW there is also an implementation of the distributed `redlock` algorithm available: https://www.npmjs.com/package/redlock

I know that wolkenkit ties to less external dependencies possible but this solution may be also very robust and provides already all the needed semantics for wolkenkit.

package.json
```diff
@@ -149,6 +149,7 @@
     "processenv": "1.1.0",
     "pump": "3.0.0",
     "recursive-readdir": "2.2.2",
+    "redis": "^2.8.0",
```
goloroden: We don't use `^`-ed versions, but rely on strict version pinning, to make installs more reliable and updates explicit. Could you please remove the `^` here?

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
+'use strict';
+
+const noop = require('lodash/noop');
```
goloroden: This line and the following ones should be one single block. The logic is always: A single block for integrated modules, one for npm modules, and one for local files. For each block only use a single `const`, and separate them by empty lines.

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+
+const limitAlphanumeric = require('limit-alphanumeric'),
+      mysql = require('mysql2/promise'),
+      retry = require('async-retry');
+
+const sortObjectKeys = require('../sortObjectKeys');
+
+// Max mysql/mariadb timestamp is 9999-12-31 23:59:59
```
goloroden: In comments we usually try to use normal grammar, so this should be `MySQL / MariaDB` (spelling, and spaces around the slash), and maybe the comment should have a trailing `.`. Respectively, since this is the MariaDB lock store, I'd suggest to get rid of the MySQL comment entirely.

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+      throw new Error('Password is missing.');
+    }
+    if (!database) {
+      throw new Error('Database is missing.');
+    }
+    if (!namespace) {
+      throw new Error('Namespace is missing.');
+    }
+
+    this.namespace = `store_${limitAlphanumeric(namespace)}`;
```
goloroden: The `store_` prefix is the same that we use in the event store. For now I'd suggest to change this to maybe `lockstore_`, to separate both from each other. In the long term, I think that we will get rid of this prefixes entirely, but hand in the table names. However, for now I'd suggest to keep this different from the event store. The same is true for all other adapters of the lock store,

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+      });
+      connection.on('end', Lockstore.onUnexpectedClose);
+    });
+
+    const connection = await this.getDatabase();
+
+    await connection.query(`
+      CREATE TABLE IF NOT EXISTS ${this.namespace}_locks (
+        namespace VARCHAR(64) NOT NULL,
+        -- 2713 is 2712bytes + 1byte for mysql varchar
```
goloroden: I don't know if I would try to mimic one database's behavior in another one, too. The question then is: Which one do you use as reference? I think that it's not very probable that people will switch databases in the running system, so why not allow what the database can offer, and then live with it?

Apart from that, I'd suggest moving the comments out of the SQL, as a JavaScript comment above the function call. This makes it more obvious that it is a comment.

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+
+    const connection = await this.getDatabase();
+
+    await connection.query(`
+      CREATE TABLE IF NOT EXISTS ${this.namespace}_locks (
+        namespace VARCHAR(64) NOT NULL,
+        -- 2713 is 2712bytes + 1byte for mysql varchar 
+        -- to match postgres max index entry size
+        value VARCHAR(2713) NOT NULL, 
+        -- bigint allows to avoid datetime/timestamp restrictions
```
goloroden: Maybe I'm missing something here, but you don't use `bigint`, do you?

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+        SELECT expiresAt
+          FROM ${this.namespace}_locks
+          WHERE namespace = ?
+            AND value = ?`,
+      [ namespace, sortedSerializedValue ]);
+
+      let newEntry = true;
+
+      if (rows.length > 0) {
+        const entry = rows[0];
```
goloroden: I think this should use destructuring (IMHO the linter also should complain about this).

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+            AND value = ?
+        `;
+      }
+
+      await connection.query(query, [ namespace, sortedSerializedValue, new Date(expiresAt) ]);
+
+      try {
+        await onAcquired();
+      } catch (ex) {
+        await this.releaseLock({
```
goloroden: This line wouldn't be that long it put into a single line, so maybe it makes sense to change this.

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  async isLocked ({ namespace, value }) {
+    if (!namespace) {
+      throw new Error('Namespace is missing.');
+    }
+    if (!value) {
+      throw new Error('Value is missing.');
+    }
+    const sortedSerializedValue = JSON.stringify(sortObjectKeys(value, true));
```
goloroden: Here, an empty line is missing.

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+    try {
+      const [ rows ] = await connection.query(`
+        SELECT expiresAt
+          FROM ${this.namespace}_locks
+          WHERE namespace = ?
+            AND value = ?`,
+      [ namespace, sortedSerializedValue ]);
+
+      if (rows.length > 0) {
+        const entry = rows[0];
```
goloroden: Again, destructuring…

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+
+        isLocked = entry.expiresAt.getTime() > Date.now();
```
goloroden: In the in-memory implementation I tried to avoid `>` comparisons for time, so that you always read from left to right with the semantics `earlier < later`. I think it makes sense to apply this here as well.

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+        SELECT expiresAt
+          FROM ${this.namespace}_locks
+          WHERE namespace = ?
+            AND value = ?`,
+      [ namespace, sortedSerializedValue ]);
+
+      if (rows.length === 0) {
+        throw new Error('Failed to renew lock.');
+      } else {
+        const entry = rows[0];
```
goloroden: Destructuring … 😉

stores/lockstore/MariaDb/Lockstore.js
```diff
@@ -0,0 +1,259 @@
...
+      const [ rows ] = await connection.query(`
+        SELECT expiresAt
+          FROM ${this.namespace}_locks
+          WHERE namespace = ?
+            AND value = ?`,
+      [ namespace, sortedSerializedValue ]);
+
+      if (rows.length === 0) {
+        throw new Error('Failed to renew lock.');
+      } else {
```
goloroden: We don't need the `else` here, as the `throw` will exit the `try` block anyway.

stores/lockstore/MongoDb/Lockstore.js
```diff
@@ -0,0 +1,205 @@
+'use strict';
+
+const noop = require('lodash/noop');
```
goloroden: This line needs to be merged with the third block, and the `url` loading line must come first.

stores/lockstore/MongoDb/Lockstore.js
```diff
@@ -0,0 +1,205 @@
...
+      throw new Error('Password is missing.');
+    }
+    if (!database) {
+      throw new Error('Database is missing.');
+    }
+    if (!namespace) {
+      throw new Error('Namespace is missing.');
+    }
+
+    this.namespace = `store_${limitAlphanumeric(namespace)}`;
```
goloroden: See above.

stores/lockstore/MongoDb/Lockstore.js
```diff
@@ -0,0 +1,205 @@
...
+
+    const sortedSerializedValue = JSON.stringify(sortObjectKeys(value, true));
+    const query = {
+      namespace,
+      value: sortedSerializedValue
+    };
+    const entry = await this.collections.locks.findOne(query);
+
+    if (entry) {
+      const isLocked = entry.expiresAt.getTime() > Date.now();
```
goloroden: `<` instead of `>`… see above.

stores/lockstore/MongoDb/Lockstore.js
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  async isLocked ({ namespace, value }) {
+    if (!namespace) {
+      throw new Error('Namespace is missing.');
+    }
+    if (!value) {
+      throw new Error('Value is missing.');
+    }
+    const sortedSerializedValue = JSON.stringify(sortObjectKeys(value, true));
```
goloroden: Here an empty line is missing.

stores/lockstore/MongoDb/Lockstore.js
```diff
@@ -0,0 +1,205 @@
...
+    const query = {
+      namespace,
+      value: sortedSerializedValue
+    };
+    const entry = await this.collections.locks.findOne(query);
+
+    const isLocked = Boolean(entry) && entry.expiresAt.getTime() > Date.now();
```
goloroden: `<` instead of `>`… see above.

stores/lockstore/MySql/Lockstore.js
```diff
@@ -0,0 +1,260 @@
...
+
+const noop = require('lodash/noop');
+
+const limitAlphanumeric = require('limit-alphanumeric'),
+      mysql = require('mysql2/promise'),
+      retry = require('async-retry');
+
+const sortObjectKeys = require('../sortObjectKeys');
+
+// Max mysql/mariadb timestamp is 9999-12-31 23:59:59
```
goloroden: Same as with the MariaDB adapter. See comment there 😉

stores/lockstore/MySql/Lockstore.js
```diff
@@ -0,0 +1,260 @@
+'use strict';
+
+const noop = require('lodash/noop');
```
goloroden: Again, this line should be part of the next block.

stores/lockstore/MySql/Lockstore.js
```diff
@@ -0,0 +1,260 @@
...
+      throw new Error('Password is missing.');
+    }
+    if (!database) {
+      throw new Error('Database is missing.');
+    }
+    if (!namespace) {
+      throw new Error('Namespace is missing.');
+    }
+
+    this.namespace = `store_${limitAlphanumeric(namespace)}`;
```
goloroden: Prefix… 😉

stores/lockstore/MySql/Lockstore.js
```diff
@@ -0,0 +1,260 @@
...
+      });
+      connection.on('end', Lockstore.onUnexpectedClose);
+    });
+
+    const connection = await this.getDatabase();
+
+    await connection.query(`
+      CREATE TABLE IF NOT EXISTS ${this.namespace}_locks (
+        namespace VARCHAR(64) NOT NULL,
+        -- 2713 is 2712bytes + 1byte for mysql varchar
```
goloroden: See comment in MariaDB adapter.

stores/lockstore/MySql/Lockstore.js
```diff
@@ -0,0 +1,260 @@
...
+        SELECT expiresAt
+          FROM ${this.namespace}_locks
+          WHERE namespace = ?
+            AND value = ?`,
+      [ namespace, sortedSerializedValue ]);
+
+      let newEntry = true;
+
+      if (rows.length > 0) {
+        const entry = rows[0];
```
goloroden: Destructuring … 😉

stores/lockstore/MySql/Lockstore.js
```diff
@@ -0,0 +1,260 @@
...
+        const isLocked = entry.expiresAt.getTime() > Date.now();
```
goloroden: `<` instead of `>`… see above.

stores/lockstore/MySql/Lockstore.js
```diff
@@ -0,0 +1,260 @@
...
+            AND value = ?
+        `;
+      }
+
+      await connection.query(query, [ namespace, sortedSerializedValue, new Date(expiresAt) ]);
+
+      try {
+        await onAcquired();
+      } catch (ex) {
+        await this.releaseLock({
```
goloroden: Again, maybe we could put this into a single line.

stores/lockstore/MySql/Lockstore.js
```diff
@@ -0,0 +1,260 @@
...
+    try {
+      const [ rows ] = await connection.query(`
+        SELECT expiresAt
+          FROM ${this.namespace}_locks
+          WHERE namespace = ?
+            AND value = ?`,
+      [ namespace, sortedSerializedValue ]);
+
+      if (rows.length > 0) {
+        const entry = rows[0];
```
goloroden: Destructuring 😉

stores/lockstore/Postgres/Lockstore.js
```diff
@@ -0,0 +1,312 @@
+'use strict';
+
+const noop = require('lodash/noop');
```
goloroden: This line should be part of the next block.

stores/lockstore/Postgres/Lockstore.js
```diff
@@ -0,0 +1,312 @@
...
+      throw new Error('Password is missing.');
+    }
+    if (!database) {
+      throw new Error('Database is missing.');
+    }
+    if (!namespace) {
+      throw new Error('Namespace is missing.');
+    }
+
+    this.namespace = `store_${limitAlphanumeric(namespace)}`;
```
goloroden: Prefix 😉

stores/lockstore/Postgres/Lockstore.js
```diff
@@ -0,0 +1,312 @@
...
+          WHERE "namespace" = $1
+            AND "value" = $2
+      `,
+        values: [ namespace, JSON.stringify(value) ]
+      });
+
+      let newEntry = true;
+
+      if (result.rows.length > 0) {
+        const entry = result.rows[0];
```
goloroden: Destructuring 😉

stores/lockstore/Postgres/Lockstore.js
```diff
@@ -0,0 +1,312 @@
...
+
+        const isLocked = entry.expiresAt.getTime() > Date.now();
```
goloroden: `<` instead of `>`… see above.

stores/lockstore/Postgres/Lockstore.js
```diff
@@ -0,0 +1,312 @@
...
+      await connection.query({
+        name: `acquire ${newEntry ? 'new' : 'existing'} lock`,
+        text: query,
+        values: [ namespace, JSON.stringify(value), new Date(expiresAt) ]
+      });
+
+      try {
+        await onAcquired();
+      } catch (ex) {
+        await this.releaseLock({
```
goloroden: Single line 😉

stores/lockstore/Postgres/Lockstore.js
```diff
@@ -0,0 +1,312 @@
...
+          WHERE "namespace" = $1
+            AND "value" = $2
+      `,
+        values: [ namespace, JSON.stringify(value) ]
+      });
+
+      if (result.rows.length > 0) {
+        const entry = result.rows[0];
+
+        isLocked = entry.expiresAt.getTime() > Date.now();
```
goloroden: `<` instead of `>`… see above.

stores/lockstore/Postgres/Lockstore.js
```diff
@@ -0,0 +1,312 @@
...
+        SELECT "expiresAt"
+          FROM "${this.namespace}_locks"
+          WHERE "namespace" = $1
+            AND "value" = $2
+      `,
+        values: [ namespace, JSON.stringify(value) ]
+      });
+
+      if (result.rows.length > 0) {
+        const entry = result.rows[0];
```
goloroden: Destructuring 😉

stores/lockstore/Postgres/Lockstore.js
```diff
@@ -0,0 +1,312 @@
...
+          WHERE "namespace" = $1
+            AND "value" = $2
+      `,
+        values: [ namespace, JSON.stringify(value) ]
+      });
+
+      if (result.rows.length === 0) {
+        throw new Error('Failed to renew lock.');
+      } else {
+        const entry = result.rows[0];
```
goloroden: Destructuring 😉

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
+'use strict';
+
+const noop = require('lodash/noop');
```
goloroden: This line should be part of the next block.

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+    }
+    if (!database) {
+      throw new Error('Database is missing.');
+    }
+    if (!namespace) {
+      throw new Error('Namespace is missing.');
+    }
+
+    this.username = username;
+    this.namespace = `store_${limitAlphanumeric(namespace)}`;
```
goloroden: Prefix 😉

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+    this.requireValidExpiration = requireValidExpiration;
+
+    const url = `redis://:${password}@${hostname}:${port}/${database}`;
```
goloroden: Maybe I'm missing something here, or maybe this is just a strange syntax, but isn't the username missing here?

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+
+    this.client = await retry(() => new Promise((resolve, reject) => {
+      const client = redis.createClient({ url });
+
+      client.ping(err => {
+        if (err) {
+          reject(err);
```
goloroden: This must  be `return reject(err);`, otherwise the `resolve` runs in case of errors, too.

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+      if (this.requireValidExpiration) {
+        throw new Error('Redis cannot acquire a lock in the past.');
+      }
+
+      result = 'OK';
+    } else {
+      result = await new Promise((resolve, reject) => {
+        this.client.set(key, this.username, 'NX', 'PX', expiration, (err, reply) => {
+          if (err) {
+            reject(err);
```
goloroden: A `return` is missing here.

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  async isLocked ({ namespace, value }) {
+    if (!namespace) {
+      throw new Error('Namespace is missing.');
+    }
+    if (!value) {
+      throw new Error('Value is missing.');
+    }
+    const key = Lockstore.getLockName({ namespace, value, store: this.namespace });
```
goloroden: Here an empty line is missing.

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+
+    const existingLock = await new Promise((resolve, reject) => {
+      this.client.get(key, (err, reply) => {
+        if (err) {
+          reject(err);
```
goloroden: `return …` … 😉

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+      if (this.requireValidExpiration) {
+        throw new Error('Redis cannot renew a lock in the past.');
+      }
+
+      result = 'OK';
+    } else {
+      const existingLock = await new Promise((resolve, reject) => {
+        this.client.get(key, (err, reply) => {
+          if (err) {
+            reject(err);
```
goloroden: Again, the `return` is missing.

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+
+          resolve(reply);
+        });
+      });
+
+      if (existingLock === this.username) {
+        result = await new Promise((resolve, reject) => {
+          this.client.pexpire(key, expiration, (err, reply) => {
+            if (err) {
+              reject(err);
```
goloroden: Here, the `return` is missing.

stores/lockstore/Redis/Lockstore.js
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    const key = Lockstore.getLockName({ namespace, value, store: this.namespace });
+
+    let result;
+
+    const existingLock = await new Promise((resolve, reject) => {
+      this.client.get(key, (err, reply) => {
+        if (err) {
+          reject(err);
```
goloroden: `return`… 😉

stores/lockstore/SqlServer/Lockstore.js
```diff
@@ -0,0 +1,391 @@
+'use strict';
+
+const noop = require('lodash/noop');
```
goloroden: This line should be part of the next block.

stores/lockstore/SqlServer/Lockstore.js
```diff
@@ -0,0 +1,391 @@
...
+
+const limitAlphanumeric = require('limit-alphanumeric'),
+      { Request, TYPES } = require('tedious'),
+      retry = require('async-retry');
+
+const createPool = require('./createPool'),
+      sortObjectKeys = require('../sortObjectKeys');
+
+// Max sqlserver datetime2 is 9999-12-31 23:59:59
```
goloroden: `sqlserver` => `SQL Server`, and a `.` in the end of the line would be nice 😊

stores/lockstore/SqlServer/Lockstore.js
```diff
@@ -0,0 +1,391 @@
...
+    const connection = await this.getDatabase();
+
+    const query = `
+      BEGIN TRANSACTION setupTable;
+
+      IF NOT EXISTS (SELECT [name] FROM sys.tables WHERE [name] = '${this.namespace}_locks')
+        BEGIN
+          CREATE TABLE [${this.namespace}_locks] (
+            [namespace] NVARCHAR(64) NOT NULL,
+            [value] NVARCHAR(2713) NOT NULL,
```
goloroden: I assume this is the `2713` coming from PostgreSQL again. As said before, I wouldn't apply other DBMS's constraints here.

stores/lockstore/SqlServer/Lockstore.js
```diff
@@ -0,0 +1,391 @@
...
+          lockResult = {
+            expiresAt: columns[0].value
+          };
+        });
+
+        database.execSql(request);
+      });
+
+      if (result) {
+        isLocked = result.expiresAt.getTime() > Date.now();
```
goloroden: `<` instead of `>`… see above.

stores/lockstore/sortObjectKeys.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const sortObjectKeys = function (object, recursive = false) {
```
goloroden: Maybe it would make sense to introduce an options object here. We generally do this for most functions, especially if things such as boolean parameters are given. For the caller it's hard to tell what `true` as second parameter means. If instead one could write

```javascript
sortObjectKeys({ object: foo, recursive: true });
```

things become clearer.

test/shared/containers/getConnectionOptions.js
```diff
@@ -50,6 +50,14 @@ const getConnectionOptions = function () {
       username: 'wolkenkit',
       password: 'wolkenkit'
     },
+    redis: {
+      hostname: 'localhost',
+      port: 6379,
+      username: 'wolkenkit',
+      password: 'wolkenkit',
+      database: '0',
```
goloroden: Does this have to be `'0'` for a specific reason, or can we choose the name freely? If so, maybe we could use something more readable than just a 0.

test/shared/containers/redis.js
```diff
@@ -0,0 +1,61 @@
...
+      port,
+      password
+    } = connectionOptions.redis;
+
+    shell.exec(oneLine`
+      docker run
+        -d
+        -p ${port}:6379
+        --name test-redis
+        thenativeweb/wolkenkit-redis:latest
```
goloroden: Again, what about the username here?

test/unit/stores/lockstore/sortObjectKeysTests.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+const assert = require('assertthat');
+
+const sortObjectKeys = require('../../../../stores/lockstore/sortObjectKeys');
+
+suite('sortObjectKeys', () => {
+  suite('default behavior', () => {
+    test('does not sort object keys', async () => {
```
goloroden: The test description is missing a `.` in the end. Did you run the linter (I'm asking because the linter should complain about this, if it doesn't, then something seems to be misconfigured)?

test/unit/stores/lockstore/sortObjectKeysTests.js
```diff
@@ -0,0 +1,70 @@
...
+
+      assert.that(actual).is.equalTo(expected);
+    });
+  });
+
+  test('is a function.', async () => {
+    assert.that(sortObjectKeys).is.ofType('function');
+  });
+
+  test('returns an array as it is', async () => {
```
goloroden: Missing trailing `.` at the test description.

test/unit/stores/lockstore/sortObjectKeysTests.js
```diff
@@ -0,0 +1,70 @@
...
+    const value = sortObjectKeys([ 3, 2, 1 ]);
+
+    const expected = '[3,2,1]';
+    const actual = JSON.stringify(value);
+
+    assert.that(actual).is.equalTo(expected);
+  });
+
+  test('returns any other value as it is', async () => {
```
goloroden: Missing trailing `.` at the test description.

stores/lockstore/MariaDb/Lockstore.js
damienbenon: This is indeed the maximum size of an index entry with Postgres.

I would like to discuss on this one: I don't think it is a good idea to let users the ability to lock too big values as you will pay it strongly performance wise, especially on the network. 

As you have to provide a size for the `value` column in sql databases, that made sense to me to have a consistent behavior across all of them by taking the lowest high limit of supported systems.

What do you think?

goloroden: I agree on that 👍 

However, I wouldn't use PostgreSQL as reference. If in the future we support another database, whose maximum value is maybe less, then we'd need to adjust this again. And, 2713 bytes seems to be somewhat arbitrary 😉 

What about limiting it to 2048 bytes? This at least looks like a well-defined boundary (2k).

Would this be fine for you?

stores/lockstore/MariaDb/Lockstore.js
damienbenon: I usually just apply the automatic fixes. I can't see any warning using IntelliJ or VSCode about that. Will fix that.

goloroden: If you run `npx roboter` instead of `npx roboter test`, our ESLint rules are also all applied.

stores/lockstore/MariaDb/Lockstore.js
damienbenon: sounds good! 2048 bytes is more than enough.

At the moment I haven't applied this restriction on Mongo nor Redis. Shall I introduce it also in those systems?

goloroden: For the sake of consistency, and to show that this is by intention, and not just a random value, I'd say "yes".

stores/lockstore/MariaDb/Lockstore.js
damienbenon: Will use that instead :)

stores/lockstore/Redis/Lockstore.js
damienbenon: Oh yeah I missed that! thanks

stores/lockstore/Redis/Lockstore.js
damienbenon: Redis does not work like a classic database. 

Username is not required and you can configure optional authentication, which will only challenge your for the password.

test/shared/containers/getConnectionOptions.js
damienbenon: Redis does not work like a classic database. You cannot create your own schema/database like you would do in a classic RDBMS.

It only provide logical databases identified by a 0-based index. Redis has by default 16 databases. You can see them somehow as memory partitions.

stores/lockstore/sortObjectKeys.js
damienbenon: sure!

test/shared/containers/getConnectionOptions.js
goloroden: Okay, I got that, thanks for the explanation 😊

damienbenon:
Hi @goloroden, I've updated the pull request with your comments.

I've also included, as discussed, a way to control the maximum size of a locked value. This is now provided with an additional parameter during `Lockstore` initialization.

2 systems have pretty low hard limits that led to have this max lock size configurable:
* MongoDb max index entry is 1024 bytes
* SqlServer clustered index (applied on PK) is 900 bytes. This is even worst as both `namespace` and `value` are counted, thus a longer namespace will reduce the maximum value size.

Also I've slightly updated the `Commandstore` to inject the `Lockstore` as now all implementation require this new `maxLockSize` setting.

goloroden:
Thanks a lot, this really looks great! Now, before merging, there is only one question open from my side:

> In your PR description, you said that you are using the username for Redis as well. To be honest, I didn't get why, respectively I didn't get why this is not needed for the other implementations?

damienbenon:
This is to ensure lock ownership.

By using the username as the value in Redis, that allows us to ensure that the lock belongs to this instance of lockstore. 

We limit then the capability of other instances to hijack a lock because for some exotic reason we try to renew/release a lock. This is part of the base implementation of the distributed lock (https://redis.io/topics/distlock#correct-implementation-with-a-single-instance).

Maybe then `username` is not appropriate, rather `instanceName`...
This could be also applied on other implementations...

goloroden:
Thanks for the explanation. I think the most relevant part of the documentation you linked is this paragraph:

> This is important in order to avoid removing a lock that was created by another client. For example a client may acquire the lock, get blocked in some operation for longer than the lock validity time (the time at which the key will expire), and later remove the lock, that was already acquired by some other client. Using just DEL is not safe as a client may remove the lock of another client. With the above script instead every lock is “signed” with a random string, so the lock will be removed only if it is still the one that was set by the client trying to remove it.

Sounds reasonable, but this OTOH means that right now the Redis-based lockstore behaves differently than all the other ones, doesn’t it?

damienbenon:
> Sounds reasonable, but this OTOH means that right now the Redis-based lockstore behaves differently than all the other ones, doesn’t it?

Yes it behaves slightly differently. Same could be quite easily done for other stores by adding an `instanceId` or `nonce` column to store the client id.

But then we would have to hack this security somehow if end users are gonna use this `Lockstore` like an unique index as they would then expect any lockstore client in the cluster to control all the values in the index...

goloroden:
After having slept over it, I still have a number of questions, but I will contact you in Slack to avoid lengthy discussions in this PR.

damienbenon:
As per our discussion on Wolkenkit's Slack, here are the changes for managing strict locking on each distributed store like `Redis` implementation already proposed.

`InMemory` implementation does not support the feature as it is not distributed per nature.

All implementations support the default unit and integration test suites.

If `nonce` is provided as parameter to the `initialize` method, then strict lock mode is automatically enabled and the underlying lockstore implementation will validate the lock belonging on `renewLock` and `releaseLock` operations.

`renewLock` `renews the lock.` integration test delays have been slightly adjusted to be more strict as `SqlServer` implementation tends to be flaky on a pretty correct machine. Maybe `SqlServer` implementation should not be proposed? Limitations are already pretty hard on `SqlServer` as per the 900 bytes limit on an index key and response times are not fast. If used on Azure, there is already a managed `Redis` service that would naturally superseed the `SqlServer` lockstore implementation...

Injection of `Lockstore` into `Commandstore` has been removed as per your request.

`createPool` util for `SqlServer` has been extracted to be reused in both `Eventstore` and `Lockstore` implementations.

PR #163, opened 2019-07-21
Switch to TypeScript.

goloroden:


tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "declaration": true,
+    "esModuleInterop": true,
+    "module": "commonjs",
+    "outDir": "build",
+    "resolveJsonModule": true,
+    "rootDir": ".",
+    "strict": true,
+    "target": "es2019"
```
gossi: I'm better off, with using `paths` here and give explicit locations for various namespaces. Also helpful for (based on ember-cli-typescript default) is to have a `types/` folder, that only contain `*.d.ts` files to type external libs and put the `types/` folder into `paths` as well for type resolutions of those.

common/utils/http/ClientMetadata.ts
```diff
@@ -0,0 +1,22 @@
+class ClientMetadata {
+  public token: string;
```
gossi: `public` is the default and not necessarily needed (configurable via *lint). We don't use it, but can't find the article I recently read the describes this as an anti-pattern. Just leaving this thoughts here. In the end it is about being consistent within one project and stick with it. Also a bit personal preference I guess.

I'm fine with either way (with java done back in the days it feels natural to put this keyword here).

global.d.ts
```diff
@@ -0,0 +1,21 @@
+declare module 'defekt' {
```
gossi: with a `types/` folder this can all go into individual files:

- `types/defekt.d.ts`
- `types/flaschenpost.d.ts`
- ...

This keeps it clean. Also with individual files `declare module` is not required.

tsconfig.json
gossi: Especially very helpful is (things I was missing when writing types for wolkenkit at first), is to define its own packages and namespaces. This is extremly supportive in getting a clean architecture done, for you own code as well as the code for consumers (and maybe to organize the monorepo based on those packages, `yarn workspaces` ftw!).

- `@wolkenkit/server`
- `@wolkenkit/client`

-> I have no idea if those packages would make sense with the current restructurings, take them as hint.

goloroden: I will have to dive deeper into both, defining own packages and namespaces, as well as using paths. Thanks for introducing these terms, now I have something to look for 😊

common/utils/http/ClientMetadata.ts
goloroden: Generally speaking, we prefer explicit over implicit. If you don't specify the access modifier, you have to know that `public` is the default. This way, you can see it.

global.d.ts
goloroden: Ah, okay 👍. Thanks for pointing this out!

tsconfig.json
gossi: https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping Here is the reference I forgot. Ping me in slack if you need help.

gossi: you can point `paths` with `*` as module to `types/` (see below) - very handy.

goloroden: Thanks a lot 👍

global.d.ts
goloroden: Two questions on this:

- We tried that, but if we simply remove `declare module`, the compiler is not able to resolve the module name any more. We're probably missing something here. Any hints?
- Having individual `.d.ts` files is the opposite of what is recommended by the TypeScript team (see [this paragraph](https://www.typescriptlang.org/docs/handbook/modules.html#ambient-modules) for details). Why do you prefer one over the other?

goloroden: Okay, we got it 😊

goloroden:
Hey @yeldiRium 😊 

Although we are not yet finally done with the migration to TypeScript, as discussed today we meanwhile have decided to definitely go the TypeScript way, and merging this back to the `monorepo` branch makes it easier to merge the documentation as well.

Hence, would you mind merging this PR (I'll leave it up to you if you squash right now, or if we keep this for later … I don't care)?

PR #276, opened 2019-11-16
Implement in-memory priority queue.

goloroden:
Hi @yeldiRium 👋 

As announced on Slack, I have found the time to implement a first draft of an in-memory priority queue this weekend.

There's still room for improvement (e.g., a lookup index for efficiently accessing individual elements is missing), and also I'm not perfectly sure about the API, but the basic implementation seems to work, and it is tested 😊 

Could you please have a look at it, and decide, whether you already would like to squash / merge it, or if we should take further steps first before doing this? If you decide to squash / merge, feel free to do so 😊

### DIFF
--- a/lib/common/data/PriorityQueue/PriorityQueue.ts
+++ b/lib/common/data/PriorityQueue/PriorityQueue.ts
@@ -1,9 +1,12 @@
 import { GetPriority } from './GetPriority';
+import { IsEqual } from './IsEqual';
 
 class PriorityQueue<TItem> {
   protected items: (TItem | undefined)[];
 
   protected getPriority: GetPriority<TItem>;
+
+  protected isEqual: IsEqual<TItem>;
 
   protected static getIndexOfLeftChild ({ index }: { index: number }): number {
     return (2 * index) + 1;
@@ -23,14 +26,21 @@ class PriorityQueue<TItem> {
     return (index - 2) / 2;
   }
 
-  public constructor ({ getPriority }: { getPriority: GetPriority<TItem> }) {
+  public constructor ({
+    getPriority,
+    isEqual = (leftItem, rightItem): boolean => leftItem === rightItem
+  }: {
+    getPriority: GetPriority<TItem>;
+    isEqual?: IsEqual<TItem>;
+  }) {
     this.items = [];
     this.getPriority = getPriority;
+    this.isEqual = isEqual;
   }
 
   protected getIndexOfItem ({ item }: { item: TItem }): number | undefined {
     for (const [ currentIndex, currentItem ] of this.items.entries()) {
-      if (item === currentItem) {
+      if (this.isEqual(item, currentItem!)) {
         return currentIndex;
       }
     }
--- a/lib/common/data/PriorityQueue/PriorityQueue.ts
+++ b/lib/common/data/PriorityQueue/PriorityQueue.ts
@@ -1,0 +1,161 @@
+import { GetPriority } from './GetPriority';
+
+// Hashtable als Index für effizientes Remove (um Linear Scan wegzubekommen)
+// => Key -> Liste(Array-Index)
+
+class PriorityQueue<TItem> {
+  protected items: (TItem | undefined)[];
+
+  protected getPriority: GetPriority<TItem>;
+
+  protected static getIndexOfLeftChild ({ index }: { index: number }): number {
+    return (2 * index) + 1;
+  }
+
+  protected static getIndexOfRightChild ({ index }: { index: number }): number {
+    return (2 * index) + 2;
+  }
+
+  protected static getIndexOfParent ({ index }: { index: number }): number {
+    const isLeftChild = index % 2 === 1;
+
+    if (isLeftChild) {
+      return (index - 1) / 2;
+    }
+
+    return (index - 2) / 2;
+  }
+
+  public constructor ({ getPriority }: { getPriority: GetPriority<TItem> }) {
+    this.items = [];
+    this.getPriority = getPriority;
+  }
+
+  protected getIndexOfItem ({ item }: { item: TItem }): number | undefined {
+    for (const [ currentIndex, currentItem ] of this.items.entries()) {
+      if (item === currentItem) {
+        return currentIndex;
+      }
+    }
+  }
+
+  protected swim ({ item, index }: { item: TItem; index: number }): void {
+    if (index === 0) {
+      return;
+    }
+
+    const parentIndex = PriorityQueue.getIndexOfParent({ index });
+    const parentItem = this.items[parentIndex];
+
+    const itemPriority = this.getPriority(item);
+    const parentPriority = this.getPriority(parentItem!);
+
+    if (parentPriority <= itemPriority) {
+      return;
+    }
+
+    this.items[parentIndex] = item;
+    this.items[index] = parentItem;
+
+    this.swim({ item, index: parentIndex });
+  }
+
+  protected sink ({ item, index }: { item: TItem; index: number }): void {
+    const leftChildIndex = PriorityQueue.getIndexOfLeftChild({ index });
+    const rightChildIndex = PriorityQueue.getIndexOfRightChild({ index });
+
+    if (leftChildIndex >= this.items.length) {
+      return;
+    }
+
+    const leftChildItem = this.items[leftChildIndex];
+    const rightChildItem = this.items[rightChildIndex];
+
+    const itemPriority = this.getPriority(item);
+
+    const leftChildItemPriority = leftChildItem ?
+      this.getPriority(leftChildItem) :
+      Number.MAX_SAFE_INTEGER;
+
+    const rightChildItemPriority = rightChildItem ?
+      this.getPriority(rightChildItem) :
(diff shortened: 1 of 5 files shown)


lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  public constructor ({ getPriority }: { getPriority: GetPriority<TItem> }) {
+    this.items = [];
+    this.getPriority = getPriority;
+  }
+
+  protected getIndexOfItem ({ item }: { item: TItem }): number | undefined {
+    for (const [ currentIndex, currentItem ] of this.items.entries()) {
+      if (item === currentItem) {
```
yeldiRium: This does not work for structurally identical but referentially different objects. Do we want to support those or do we want to use referential identity?

Since our intention is to nest simple queues into this, referential equality is probably enough. However, I'm not sure whether we'll want structural equality at some point.

lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,158 @@
...
+        return currentIndex;
+      }
+    }
+  }
+
+  protected swim ({ item, index }: { item: TItem; index: number }): void {
```
yeldiRium: I think this should contain a check for whether the element at index `index` in the queue is actually `item`. Just to make sure no one accidentally calls this with invalid parameters.

Also I find the name `swim` not clear enough. I needed to contrast it with `sink` to fully understand what it was meant as. I know `swim` and `sink` in heaps as `repairUp` and `repairDown`. What do you think of these names?

lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,158 @@
...
+      return;
+    }
+
+    this.items[parentIndex] = item;
+    this.items[index] = parentItem;
+
+    this.swim({ item, index: parentIndex });
+  }
+
+  protected sink ({ item, index }: { item: TItem; index: number }): void {
```
yeldiRium: Both points of the comment from the method `swim` apply here aswell.

lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,158 @@
+import { GetPriority } from './GetPriority';
+
+class PriorityQueue<TItem> {
```
yeldiRium: I would include a comment that indicates that this is a min-heap implementation and that lower priority values are closer to the queue's start.

lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  public async enqueue ({ item }: { item: TItem }): Promise<void> {
+    const enqueueIndex = this.items.length;
+
+    this.items[enqueueIndex] = item;
+    this.swim({ item, index: enqueueIndex });
+  }
+
+  public async dequeue ({ item }: { item: TItem }): Promise<void> {
```
yeldiRium: I would expect `dequeue` to return the first item in the queue and repair the heap. Maybe this could be named `remove` or something along the lines, while `dequeue` is `getNextItem` plus removing the first item in the queue.

lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,158 @@
...
+    if (index >= this.items.length) {
+      return;
+    }
+
+    this.items[index] = lastItem;
+
+    const itemPriority = this.getPriority(item);
+    const lastItemPriority = this.getPriority(lastItem!);
+
+    if (lastItemPriority < itemPriority) {
```
yeldiRium: If we use the last item in the priority queue to replace another item, the replaced item will either be on the same heap level or on a level above it. Therefore calling `sink` should be enough, or am I missing something?

lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  public async isEmpty (): Promise<boolean> {
+    return this.items.length === 0;
+  }
+
+  public async getNextItem (): Promise<TItem | undefined> {
+    return this.items[0];
+  }
+
+  public async getHeap (): Promise<(TItem | undefined)[]> {
```
yeldiRium: In line with `Array` and `Object` I suggest calling this method `values()`.

lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,158 @@
...
+    const lastItemPriority = this.getPriority(lastItem!);
+
+    if (lastItemPriority < itemPriority) {
+      this.swim({ item: lastItem!, index });
+    } else {
+      this.sink({ item: lastItem!, index });
+    }
+  }
+
+  public async rebalance ({ item }: { item: TItem }): Promise<void> {
```
yeldiRium: Maybe we should require `index` as a second parameter here as well, so that the interface is congruent to `swim` and `sink`. That would probably require a validation as mentioned above, but would improve this methods complexity from `O(n)` to `O(log(n))`.

lib/common/data/PriorityQueue/PriorityQueue.ts
goloroden: Actually, my first implementation included the option to include an `IsEqual` function via the constructor. Then I thought that this would probably not be needed, but we can implement this as an optional one, with the default going to be `===`. I'm going to change this.

lib/common/data/PriorityQueue/PriorityQueue.ts
goloroden: Actually, while researching I found a few variations of naming these functions, and I liked `swim` and `sink`, as they are a good-matching metaphor. But I agree that `repairUp` and `repairDown` are technically more correct, and probably better understandable if you don't have the initial context. So I'm going to rename these.

Regarding the question whether to check for `index` here… I've skipped this, since this is only an internal (i.e. `protected`) function, and it's not possible to hand over wrong values from the *outside*. From the inside I thought it would be fine like this, because it avoids another lookup.

Maybe we should change this when introducing a lookup index?

lib/common/data/PriorityQueue/PriorityQueue.ts
goloroden: And it's the same answer here, so I'm going to resolve this one with a reference to the answer given above 😉

lib/common/data/PriorityQueue/PriorityQueue.ts
goloroden: Done 👍

lib/common/data/PriorityQueue/PriorityQueue.ts
goloroden: I like this idea 👍

lib/common/data/PriorityQueue/PriorityQueue.ts
goloroden: Same argument as above: This is a public function, and I think that from the outside you shouldn't have to care about that there is something such as an index – in contrast to `swim`/`sink` which are only for internal usage. What do you think?

lib/common/data/PriorityQueue/PriorityQueue.ts
goloroden: That's actually true. The last item's priority must be greater than or equal to the one being removed, hence it can only sink, not swim. I'm going to fix this.

lib/common/data/PriorityQueue/PriorityQueue.ts
goloroden: That's true, and actually I had the same thought in the beginning, but canceled it for some reason I have forgotten 😉 

I'm going to change this.

lib/common/data/PriorityQueue/PriorityQueue.ts
```diff
@@ -0,0 +1,175 @@
...
+    if (isLeftChild) {
+      return (index - 1) / 2;
+    }
+
+    return (index - 2) / 2;
+  }
+
+  public constructor ({
+    getPriority,
+    isEqual = (leftItem, rightItem): boolean => leftItem === rightItem
```
yeldiRium: I like this solution :+1:

lib/common/data/PriorityQueue/PriorityQueue.ts
yeldiRium: Ah, I completely overlooked the visibility modifiers. Yes, for protected methods skipping the check is fine.

Regarding the lookup index: I don't think we'll need the check when we have a separate index, since access to the objects still doesn't allow removing them from the priority queue. Rebalancing will be another problem, but let's discuss that when implementing the lookup index.

lib/common/data/PriorityQueue/PriorityQueue.ts
yeldiRium: You are right about hiding the index from the public. I dislike that this makes the operation take longer on average, but we'll have to see if that is actually a problem. We can keep it like this for now.

PR #277, opened 2019-11-17
Turn asynchronous functions to synchronous ones.

goloroden:
Hey @yeldiRium 😊 

As announced on Slack, I think it's a) a good idea to keep the priority queue generic, and b) to turn the asynchronous functions to synchronous ones.

For a) we didn't need to do anything, for b) I have created this PR.

If you're fine with it, can you please squash / merge it?

### DIFF
--- a/lib/common/data/PriorityQueue/PriorityQueue.ts
+++ b/lib/common/data/PriorityQueue/PriorityQueue.ts
@@ -109,26 +109,26 @@ class PriorityQueue<TItem> {
     }
   }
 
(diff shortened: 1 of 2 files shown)


yeldiRium:
As discussed we'll leave the implementation asynchronous, since we need the asynchronous interface anyways for persistent implementations.

yeldiRium:
Reopening and merging since we want to use the synchronous version as base for an asynchronous version which uses the synchronous methods internally.

PR #296, opened 2019-11-19
Try to improve get available ports.

goloroden:
Hey @yeldiRium 👋 

We *still* have this annoying problem with the `getAvailablePorts` function. I now introduced _another_ delay, so maybe this helps … if not, maybe it would make sense to re-think the entire approach.

Could you please review, squash / merge this?


goloroden:
@yeldiRium Okay, after a long evening, I think I might have a solution now that works, and I may have found the problem …

To cut a long story short, in the end I turned back to our custom-made solution, without an external module (which made things even worse), and I finally made it to remove the calls to `sleep`, which also didn't seem to help.

What seemed to actually help was to bind to `0.0.0.0` instead of `127.0.0.1` when trying a port. This makes way more sense, and TBH IDK why it was `127.0.0.1` originally. However, finally this PR is green again, so maybe it stays like this 😉 

At least we could give it a new try. Would you mind to squash / merge?

PR #314, opened 2019-11-23
Migrate back to mysql2.

goloroden:
The `mysql` module is pretty much a desaster. It's now one year and a half that work on MySQL 8 support started, and since then it didn't land. Instead, we already have the second PR (see https://github.com/mysqljs/mysql/pull/2233), which does not get merged, due to outstanding changes…

It's not even possible to offer help, or ask how one could help, since everything is locked now: You can't comment, you can't create new issues, …

Meanwhile, support for MySQL 8 has landed in `mysql2`, which we already used in earlier times. We switched away because of the missing TypeScript support, but this is also available meanwhile.

So this PR re-introduces `mysql2`, but it does not yet work perfectly, since the permission system in MySQL 8 has changed: We might need to rethink whether we actually want to setup databases on first connect, but extract this into an additional step you need to trigger manually.


goloroden:
Will be done as part of #345.

Issue #339, opened 2019-12-10
Distribute commands over multiple domain servers


goloroden:
### What is this feature about?

In the past, commands were only handled by a single domain server (back then, it was called the `core`). It would be great to be able to distribute commands over multiple domain servers, to be able to scale and to load-balance. However, one must pay attention to the fact that no two commands for the same aggregate may be run at the same time.

We will have the following routes:

- `GET /` is a route that stays open until a result is available, which means it needs to send a ping from time to time to keep the connection open. Once a command could be locked successfully, this route returns this command and gets closed by the server.
- `POST /renew-lock` renews a lock, and the required data is sent in the body as JSON.
- `POST /acknowledge` acknowledges a lock, and the required data is sent in the body as JSON.

### What needs to be done to implement this feature?

- [x] Implement an in-memory priority queue
  - [x] Add unit tests
  - [x] Use `contextIdentifier`, `aggregateIdentifier`, `id` and `name` as parameters for `renewLock` and `acknowledge`
- [x] Design a `awaitCommandWithMetadata` API
- [x] Implement the `awaitCommandWithMetadata` API
  - [x] Extract a `streamNdjsonMiddleware` (or something similar`) from the domain events API, which takes care of the heartbeat, the headers, and so on, and reuse it
  - [x] Add unit tests
- [x] Build a server for the `microservices` runtime
  - [x] Use the `handleCommandWithMetadata` API
  - [x] Use the `awaitCommandWithMetadata` API
  - [x] Add runtime tests

yeldiRium:
Done :)

PR #350, opened 2019-12-11
Rewrite priority queue store interface.

yeldiRium:
Hi @goloroden,

as in #339 discussed, I've rewritten the interface of the `PriorityQueueStore` to accept only the identifiers of the relevant items.

To help with that I've added the `ItemIdentifier` interface and placed it in the `priorityQueueStore` folder, since it is not relevant to the rest of the project so far.

Please take a look and feel free to merge, if all is well.

### DIFF
--- a/test/integration/stores/priorityQueueStore/getTestsFor.ts
+++ b/test/integration/stores/priorityQueueStore/getTestsFor.ts
@@ -8,7 +8,7 @@ import { PriorityQueueStore } from '../../../../lib/stores/priorityQueueStore/Pr
 import { sleep } from '../../../../lib/common/utils/sleep';
 import { uuid } from 'uuidv4';
 
-const itemIdentifierFromCommand = function (command: CommandWithMetadata<CommandData>): ItemIdentifier {
+const getItemIdentifierFromCommand = function (command: CommandWithMetadata<CommandData>): ItemIdentifier {
   return {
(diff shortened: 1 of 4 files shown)


test/integration/stores/priorityQueueStore/getTestsFor.ts
```diff
@@ -3,10 +3,20 @@ import { buildCommandWithMetadata } from '../../../shared/buildCommandWithMetada
 import { CommandData } from '../../../../lib/common/elements/CommandData';
 import { CommandWithMetadata } from '../../../../lib/common/elements/CommandWithMetadata';
 import { CustomError } from 'defekt';
+import { ItemIdentifier } from '../../../../lib/stores/priorityQueueStore/ItemIdentifier';
 import { PriorityQueueStore } from '../../../../lib/stores/priorityQueueStore/PriorityQueueStore';
 import { sleep } from '../../../../lib/common/utils/sleep';
 import { uuid } from 'uuidv4';
 
+const itemIdentifierFromCommand = function (command: CommandWithMetadata<CommandData>): ItemIdentifier {
```
goloroden: First of all: I really love the idea of having an `ItemIdentifier` type here, that's a great thing 👍 

The only thing that I think I'd change is the name of this function, to something like `get…`. So, if we just prefix this method with `get`, I'd be fine with it. Would you mind changing that?

PR #351, opened 2019-12-11
Extract streamNdjson middleware and use it in observeDomainEvents api.

yeldiRium:
Hi @goloroden,

as discussed in #339 I've extracted a middleware from the `observeDomainEvents` api that sets up a connection as using `application/x-ndjson` and sends a heartbeat.

I've added tests for the middleware and had to make slight changes to the helper `WriteStream` `asJsonStream` which luckily don't affect the other places in which the helper is used.

Feel free to merge the changes if everything is well.

### DIFF
--- a/lib/apis/middlewares/streamNdjson.ts
+++ b/lib/apis/middlewares/streamNdjson.ts
@@ -6,11 +6,9 @@ const logger = flaschenpost.getLogger();
 
 const heartbeat = { name: 'heartbeat' };
 
-/**
- * The streamNdjson middleware initializes a long-lived connection with the
- * content type `application/x-ndjson` and sends a periodic heartbeat every
- * `heartbeatInterval` milliseconds.
- */
+// The streamNdjson middleware initializes a long-lived connection with the
+// content type `application/x-ndjson` and sends a periodic heartbeat every
+// `heartbeatInterval` milliseconds.
 const streamNdjsonMiddleware = function ({
   heartbeatInterval
 }: {
--- a/lib/apis/middlewares/streamNdjson.ts
+++ b/lib/apis/middlewares/streamNdjson.ts
@@ -1,6 +1,6 @@
 import { flaschenpost } from 'flaschenpost';
-import { writeLine } from '../base/writeLine';
 import { RequestHandler } from 'express';
+import { writeLine } from '../base/writeLine';
 
 const logger = flaschenpost.getLogger();
 
--- a/lib/apis/middlewares/streamNdjson.ts
+++ b/lib/apis/middlewares/streamNdjson.ts
@@ -1,0 +1,60 @@
+import { flaschenpost } from 'flaschenpost';
+import { writeLine } from '../base/writeLine';
+import { RequestHandler } from 'express';
+
+const logger = flaschenpost.getLogger();
+
+const heartbeat = { name: 'heartbeat' };
+
+/**
+ * The streamNdjson middleware initializes a long-lived connection with the
+ * content type `application/x-ndjson` and sends a periodic heartbeat every
+ * `heartbeatInterval` milliseconds.
+ */
+const streamNdjsonMiddleware = function ({
+  heartbeatInterval
+}: {
+  heartbeatInterval: number;
+}): RequestHandler {
+  return async function (req, res, next): Promise<void> {
+    try {
+      let heartbeatIntervalId: NodeJS.Timeout;
+
+      req.setTimeout(0, (): void => undefined);
+      res.setTimeout(0);
+
+      res.writeHead(200, { 'content-type': 'application/x-ndjson' });
+
+      res.connection.once('close', (): void => {
+        clearInterval(heartbeatIntervalId);
+      });
+
+      // Send an initial heartbeat to initialize the connection. If we do not do
+      // this, sometimes the connection does not become open until the first data
+      // is sent.
+      writeLine({ res, data: heartbeat });
+
+      heartbeatIntervalId = setInterval((): void => {
+        writeLine({ res, data: heartbeat });
+      }, heartbeatInterval);
+
+      return next();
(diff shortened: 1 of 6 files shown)


lib/apis/middlewares/streamNdjson.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const logger = flaschenpost.getLogger();
+
+const heartbeat = { name: 'heartbeat' };
+
+/**
+ * The streamNdjson middleware initializes a long-lived connection with the
+ * content type `application/x-ndjson` and sends a periodic heartbeat every
+ * `heartbeatInterval` milliseconds.
+ */
```
goloroden: Can we change this to a comment using multiple `//`s? I know that we don't enforce this right now, but this is the common style used everywhere else (if we have comments), so I'd vote to stick with that.

lib/apis/middlewares/streamNdjson.ts
```diff
@@ -0,0 +1,60 @@
...
+  return async function (req, res, next): Promise<void> {
+    try {
+      let heartbeatIntervalId: NodeJS.Timeout;
+
+      req.setTimeout(0, (): void => undefined);
+      res.setTimeout(0);
+
+      res.writeHead(200, { 'content-type': 'application/x-ndjson' });
+
+      res.connection.once('close', (): void => {
```
goloroden: Is this event enough to clear things, or should we also run this code in case of errors (see the `catch` below)?

lib/apis/observeDomainEvents/http/v2/getDomainEvents.ts
```diff
@@ -68,27 +62,12 @@ const getDomainEvents = function ({
         /* eslint-enable @typescript-eslint/no-floating-promises */
       };
 
-      req.setTimeout(0, (): void => undefined);
-      res.setTimeout(0);
-
-      res.writeHead(200, { 'content-type': 'application/x-ndjson' });
-
       res.connection.once('close', (): void => {
```
goloroden: Same question here as in the middleware: Is this event enough?

test/shared/http/asJsonStream.ts
```diff
@@ -7,8 +7,9 @@ const asJsonStream = function <TItem> (...handleJson: ((item: TItem) => void)[])
     write (chunk: object, _, callback: (error?: any) => void): void {
       const data = JSON.parse(chunk.toString());
 
-      if (counter > handleJson.length) {
-        return callback(new Error(`Received ${counter + 1} items (${JSON.stringify(data)}), but only expected ${handleJson.length}.`));
+      // Ignore any messages after all handlers have been applied.
+      if (counter >= handleJson.length) {
+        return;
```
goloroden: Hmmmm 🤔 

Doesn't this stream now stay open forever, because we never close it? Is this by intention?

Oh, and: Why the change from `>` to `>=`?

test/unit/apis/middlewares/streamNdjsonTests.ts
```diff
@@ -0,0 +1,70 @@
...
+
+suite('streamNdjson middleware', (): void => {
+  let api: Application;
+
+  setup(async (): Promise<void> => {
+    api = express();
+    api.get('/', streamNdjsonMiddleware({ heartbeatInterval: 1000 }));
+  });
+
+  test('returns 200.', async (): Promise<void> => {
```
goloroden: Maybe `returns the status code 200.`?

test/unit/apis/middlewares/streamNdjsonTests.ts
```diff
@@ -0,0 +1,70 @@
...
+            }
+          ));
+        /* eslint-enable @typescript-eslint/no-floating-promises */
+      } catch (ex) {
+        reject(ex);
+      }
+    });
+  });
+
+  test('returns application/json.', async (): Promise<void> => {
```
goloroden: This is wrong, isn't it? Shouldn't this be `application/x-ndjson`?

test/shared/http/asJsonStream.ts
yeldiRium: The change from `>` to `>=` because `>` was wrong and did not produce errors in cases where it previously should.

That the stream now stays open forever is intended on my part, so that the helper can accept more stream elements without throwing an exception. The producing site should stop the producing stream. This write stream will probably stay open and produce a memory leak, but since it is only used in tests that shouldn't be a problem.

lib/apis/middlewares/streamNdjson.ts
yeldiRium: This is enough to clear things, since the interval is created in the last statement in this block and any error occuring there will result in the interval not being set. In that case it doesn't need to be cleared. Errors inside the interval can't be handled here anyway.

lib/apis/observeDomainEvents/http/v2/getDomainEvents.ts
yeldiRium: The same argument from the middleware works here. The statement that attaches a listener to the event emitter is the last one in the scope and any error occuring there will result in the cleanup not being necessary.

test/shared/http/asJsonStream.ts
goloroden: Okay, makes sense 👍

test/unit/apis/middlewares/streamNdjsonTests.ts
yeldiRium: True. I've changed it and the test still passes. Something feels wrong here. Why did the test pass in the first place? Supertest doesn't seem to work that great with streams unfortunately.

Do you have any idea on how to improve this?

goloroden: Hmmmm … good point, and yes, good question 🤔

PR #353, opened 2019-12-11
Replace supertest by custom implementation.

goloroden:
Hey @yeldiRium 👋 

As announced on Slack earlier this day, I've tried to create a replacement for `supertest`, based on `axios`. It was actually way easier than I thought, and to prove that things work I have changed the tests for the NDJSON middleware to the new component.

I have not yet replaced it everywhere, and hence I have not yet removed the dependency to `supertest`. I first wanted to get some feedback from your side.

What do you think on this approach?

(It actually solves the issues we had with the tests, since `axios` has a way better interface than `superagent` when dealing with streams.)

### DIFF
--- a/test/shared/http/runAsServer.ts
+++ b/test/shared/http/runAsServer.ts
@@ -1,0 +1,30 @@
+import { Application } from 'express';
+import { getAvailablePort } from '../../../lib/common/utils/network/getAvailablePort';
+import http from 'http';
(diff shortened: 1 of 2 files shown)


yeldiRium:
I've had a look and I like the solution a lot so far. I've played with it a bit and seems to do what it's supposed to.

I suggest we merge this PR and I replace supertest in some more places (or all places, if it works as intended).

PR #355, opened 2019-12-13
Implement awaitCommandWithMetadata API.

yeldiRium:
Hi @goloroden,

I'm done implementing the `awaitCommandWithMetadata` API.

It is constructed like most APIs and additionally requires an `applicationDefinition` (to validate the `itemIdentifiers` in `/renew-lock` and `/acknowledge` against), a `priorityQueueStore` (from which to pull the commands) and a `pollInterval` (that determains, how often commands are polled from the priority queue while connections are open).

The tests got a bit extensive, but I tried to cover as many cases as possible.

You can go ahead and merge the changes, if you're fine with them, but with a PR this large I expect something to be amiss :)


lib/apis/awaitCommandWithMetadata/http/v2/acknowledge.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (contentType.type !== 'application/json') {
+        throw new errors.RequestMalformed();
+      }
+    } catch {
+      res.status(415).send('Header content-type must be application/json.');
+
+      return;
+    }
+
+    const { itemIdentifier, token } = req.body;
```
goloroden: We do not verify the `token` in any form, so this could be malformed… would it maybe make sense to add a check for this, too? Maybe just using [this function](https://www.npmjs.com/package/uuidv4#verifying-a-uuid)?

lib/apis/awaitCommandWithMetadata/http/v2/awaitCommandWithMetadata.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+
+    const lockNextIntervalId = setInterval(async (): Promise<void> => {
+      const success = await maybeHandleLock({ priorityQueueStore, res });
+
+      if (success) {
+        clearInterval(lockNextIntervalId);
+      }
+    }, pollInterval);
```
goloroden: While this works, I think this is a problematic way to go: Supposed right now there is no command in the queue, and a worker notifies us of being idle, then we only check this every 5 seconds (by default). This means that in the worst case, if the check has just run, we need to wait 5 seconds until something happens, which is a pretty long time, if I do a very simple action in the UI.

Of course you can lower the number, but then you flood the database with the same query over and over again, run every 100ms.

Maybe a better option here would be not to pull, but to implement a notification mechanism. With the in-memory store this could be a simple event-emitter, with database-based stores we could use features such as `NOTIFY/LISTEN` (e.g., in case of PostgreSQL) or a Redis `PUBSUB` list.

Then we would have immediate command handling, and still don't run senseless queries every few ms.

What do you think?

lib/apis/awaitCommandWithMetadata/http/v2/index.ts
```diff
@@ -0,0 +1,9 @@
+import { acknowledge } from './acknowledge';
+import { awaitCommandWithMetadata } from './awaitCommandWithMetadata';
+import { renewLock } from './renewLock';
+
+export {
+  acknowledge,
+  awaitCommandWithMetadata,
```
goloroden: I wouldn't export them here in alphabetical order, but swap these two lines: The second one is the more important one, as it's the actual reason why this API exists. The first one is merely a helper.

lib/apis/awaitCommandWithMetadata/http/v2/renewLock.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (contentType.type !== 'application/json') {
+        throw new errors.RequestMalformed();
+      }
+    } catch {
+      res.status(415).send('Header content-type must be application/json.');
+
+      return;
+    }
+
+    const { itemIdentifier, token } = req.body;
```
goloroden: Same as above, maybe it would make sense to validate the token using a schema or something like that…

test/shared/getItemIdentifierFromCommand.ts
```diff
@@ -0,0 +1,16 @@
+import { CommandData } from '../../lib/common/elements/CommandData';
+import { CommandWithMetadata } from '../../lib/common/elements/CommandWithMetadata';
+import { ItemIdentifier } from '../../lib/common/elements/ItemIdentifier';
+
+const getItemIdentifierFromCommand = function (command: CommandWithMetadata<CommandData>): ItemIdentifier {
```
goloroden: Would it maybe make sense to move this function from `shared` to be a member of the `Command` and `CommandWithMetadata` classes, and simply name it `getItemIdentifier` there?

test/unit/common/validators/validateItemIdentifierTests.ts
```diff
@@ -0,0 +1,83 @@
...
+    applicationDefinition = await getApplicationDefinition({ applicationDirectory });
+  });
+
+  test('does not throw an error if everything is fine.', async (): Promise<void> => {
+    assert.that((): void => {
+      validateItemIdentifier({ itemIdentifier, applicationDefinition });
+    }).is.not.throwing();
+  });
+
+  test('throws an error if the itemIdentifier does not match the itemIdentifier schema.', async (): Promise<void> => {
```
goloroden: I'd prefer `item identifier` in the title, as this is more human readable (and in this case the test's intention is to tell something about the concept of an `item identifier`, not only technically about the implementation of the `itemIdentifier` interface).

test/unit/common/validators/validateItemIdentifierTests.ts
```diff
@@ -0,0 +1,83 @@
...
+        applicationDefinition
+      });
+    }).is.throwing(
+      (ex): boolean =>
+        (ex as CustomError).code === 'EITEMIDENTIFIERMALFORMED' &&
+        ex.message === 'String is too short (0 chars), minimum 1 (at itemIdentifier.name).'
+    );
+  });
+
+  test(`throws an error if the itemIdentifier's context doesn't exist in the application definition.`, async (): Promise<void> => {
```
goloroden: Same as above.

test/unit/common/validators/validateItemIdentifierTests.ts
```diff
@@ -0,0 +1,83 @@
...
+        applicationDefinition
+      });
+    }).is.throwing(
+      (ex): boolean =>
+        (ex as CustomError).code === 'ECONTEXTNOTFOUND' &&
+        ex.message === `Context 'someContext' not found.`
+    );
+  });
+
+  test(`throws an error if the itemIdentifier's aggregate doesn't exist in the application definition.`, async (): Promise<void> => {
```
goloroden: Once again 😉

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+        const { headers } = await client({
+          method: 'get',
+          url: '/v2/',
+          responseType: 'stream'
+        });
+
+        assert.that(headers['content-type']).is.equalTo('application/x-ndjson');
+      });
+
+      test('leaves the connection open indefiniteley as long as no command is enqueued.', async (): Promise<void> => {
```
goloroden: `indefiniteley` should be `indefinitely` (without the last `e`, I guess).

Apart from that, maybe that's a German problem, this pretty much reminds me of `nicht definiert`, instead of `unendlich` (although this is what this word means). Maybe `for an infinite amount of time`? (Although this is rather lengthy…)

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+            reject(new Error('Stream should not have closed yet.'));
+          });
+
+          data.pipe(asJsonStream(
+            (streamElement): void => {
+              assert.that(streamElement).is.equalTo({ name: 'heartbeat' });
+            }
+          ));
+
+          await sleep({ ms: 5_000 });
```
goloroden: Can we get the same effect with lower timeouts? This makes the unit (!) tests pretty slow. If it was an integration test, okay, but in unit tests I think we should try to be faster.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+
+          resolve();
+        });
+      });
+
+      test('keeps the connection open and delivers a command, then closes.', async (): Promise<void> => {
```
goloroden: We do not explicitly test that it keeps the connection open in this test. What about `closes the connection one a command has been delivered.`?

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+            (streamElement: any): void => {
+              assert.that(streamElement.item).is.equalTo(commandWithMetadata);
+              assert.that(isUuid(streamElement.token)).is.true();
+
+              resolve();
+            }
+          ));
+        });
+
+        await sleep({ ms: 5_000 });
```
goloroden: Again, it would be great if we could lower this timeout.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+            },
+            (streamElement: any): void => {
+              assert.that(streamElement.item).is.equalTo(commandWithMetadata);
+              assert.that(isUuid(streamElement.token)).is.true();
+            }
+          ));
+        });
+      });
+
+      test('delivers a locked command to the next waiting client after the lock expires.', async (): Promise<void> => {
```
goloroden: Maybe `once` instead of `after`?

(While we pulling, it's actually reasonable to use `after`, but if we turn to a notification mechanism, `once` may make more sense.)

Or: `[…] after the lock has expired.`, since maybe there is not even a connected client currently waiting.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+              assert.that(streamElement.item).is.equalTo(commandWithMetadata);
+              assert.that(isUuid(streamElement.token)).is.true();
+
+              resolve();
+            }
+          ));
+        });
+      });
+
+      test('delivers commands in different aggregates in parallel.', async (): Promise<void> => {
```
goloroden: With this test, do we really address behavior of the API, or do we test implementation behavior of the underlying queue store? IOW: Is this test really helpful?

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+          url: '/v2/renew-lock',
+          headers: { 'content-type': 'application/json' },
+          data: {
+            itemIdentifier: getItemIdentifierFromCommand(commandWithMetadata),
+            token: uuid()
+          },
+          validateStatus: (): boolean => true
+        });
+
+        assert.that(status).is.equalTo(400);
```
goloroden: Shouldn't this rather be a `404`? It's not that the request itself was bad (in terms of badly formatted), it's more like "given this token, no lock was found", isn't it?

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+        assert.that(data).is.equalTo(`Token mismatch for item 'sampleContext.sampleAggregate.${commandWithMetadata.aggregateIdentifier.id}.execute.${commandWithMetadata.id}'.`);
+      });
+
+      test('extends the lock expiry time.', async function (): Promise<void> {
+        this.timeout(15_000);
```
goloroden: This again is pretty long. Maybe we can lower it.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+            (streamElement): void => {
+              assert.that(streamElement).is.equalTo({ name: 'heartbeat' });
+            },
+            (streamElement: any): void => {
+              resolve(streamElement);
+            }
+          ));
+        });
+
+        await sleep({ ms: 3_000 });
```
goloroden: Here you rely on timing effects. First, it maybe would be a good idea to add a comment here, so that if you ever change the configured times you know how you have to configure these values here. Or we could maybe also calculate them as percentage of the configured times (which doesn't mean that it would not be a good idea to have a comment here anyway).

Similar to the comments we have in the lock store tests.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+            reject(new Error('Stream should not have closed yet.'));
+          });
+
+          unavailableLockData.pipe(asJsonStream(
+            (streamElement): void => {
+              assert.that(streamElement).is.equalTo({ name: 'heartbeat' });
+            }
+          ));
+
+          await sleep({ ms: 5_000 });
```
goloroden: Timeout … 😉

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+          url: '/v2/acknowledge',
+          headers: { 'content-type': 'application/json' },
+          data: {
+            itemIdentifier: getItemIdentifierFromCommand(commandWithMetadata),
+            token: uuid()
+          },
+          validateStatus: (): boolean => true
+        });
+
+        assert.that(status).is.equalTo(400);
```
goloroden: Same question as above …`404` instead of `400`?

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
```diff
@@ -0,0 +1,776 @@
...
+              ip: '127.0.0.1',
+              user: { id: 'jane.doe', claims: { sub: 'jane.doe' }},
+              token: '...'
+            },
+            initiator: {
+              user: { id: 'jane.doe', claims: { sub: 'jane.doe' }}
+            }
+          }
+        });
+        const commandTwo = new CommandWithMetadata({
```
goloroden: I know I'm late to the party with this comment, but didn't we have a `buildCommand` helper function one could use here (and at several other places, too)?

lib/apis/awaitCommandWithMetadata/http/v2/acknowledge.ts
yeldiRium: Yes :)

test/shared/getItemIdentifierFromCommand.ts
yeldiRium: Good idea, but also for `DomainEvent` and `DomainEventWithMetadata`, since the api is meant to be used with those aswell.

yeldiRium: Correction: We can't implement this for `Command`, since a raw `Command` has no id yet. So only for `CommandWithMetadata`, `DomainEvent` and `DomainEventWithMetadata`.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
yeldiRium: Good catch on the additional `e`. However the usage of the word is correct. It means "infinitely" or "for an undefined amount of time".

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
yeldiRium: I could lower the values used when constructing the priority queue and the API, but I experienced time fluctuations already and the current values seemed safe so far. I'll experiment with lower values.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
yeldiRium: It is helpful, since this is behavior that the API should implement, whether it is part of the underlying queue or not. If we replace the `PriorityQueue` with something else or change its implementation and this test fails, we made a mistake. I would leave the test as-is.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
yeldiRium: You're right that 400 doesn't make much sense, since the request is formatted correctly. However a 404 doesn't fit in my opinion, since a lock can be found for the `itemIdentifier`, the request just has the wrong token for it. I'd suggest a 403, since the token is an authorization mechanism.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
yeldiRium: Good catch. It slimmed down the tests a bit, so it's good you pointed it out.

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
yeldiRium: I was successful :)

test/unit/apis/awaitCommandWithMetadata/httpTests.ts
yeldiRium: I've improved the comment below and added percentage-based sleeps.

PR #370, opened 2019-12-16
feat: Implement PubSub interface and in-memory implementation.

yeldiRium:


test/integration/messaging/pubSub/InMemoryTests.ts
```diff
@@ -0,0 +1,16 @@
+import { getTestsFor } from './getTestsFor';
+import InMemoryPublisher from '../../../../lib/messaging/pubSub/InMemory/InMemoryPublisher';
+import InMemorySubscriber from '../../../../lib/messaging/pubSub/InMemory/InMemorySubscriber';
+import { Publisher } from '../../../../lib/messaging/pubSub/Publisher';
+import { Subscriber } from '../../../../lib/messaging/pubSub/Subscriber';
+
+suite.only('InMemory', (): void => {
```
goloroden: There is an `only` too much 😉

lib/messaging/pubSub/InMemory/InMemoryPubSubSingleton.ts
```diff
@@ -0,0 +1,27 @@
...
+      return this.instance;
+    }
+
+    this.instance = new InMemoryPubSubSingleton({
+      eventEmitter: new EventEmitter()
+    });
+
+    return this.instance;
+  }
+}
```
goloroden: Since this is meant to be a singleton, wouldn't it be simpler to just export an instance of the class, instead of exporting the class and implementing the singleton pattern manually?

If so, I'd also vote to rename the file to `inMemory…` (with lower case initial letter).

lib/messaging/pubSub/InMemory/InMemorySubscriber.ts
```diff
@@ -0,0 +1,28 @@
...
+    channel,
+    callback
+  }: {
+    channel: string;
+    callback: (message: T) => void | Promise<void>;
+  }): Promise<void> {
+    this.pubSubSingleton.eventEmitter.on(channel, callback);
+  }
+}
+
```
goloroden: Should we also provide an `unsubscribe` function?

lib/messaging/pubSub/InMemory/InMemoryPubSubSingleton.ts
yeldiRium: True, that is something javascript can do :joy: WILCO.

Issue #341, opened 2019-12-10
Implement an HTTP API for the event store


goloroden:
### What is this feature about?

In the past, the event store was always used directly (i.e. using a database connection) from various places, such as the read model server and the write model server. Since we intentionally crash servers if the underlying database connection goes down, this leads to the event store being a single point of failure.

To avoid this, we should introduce an HTTP API which abstracts away the actual event store interface, and which can be consumed from various places over a plain old HTTP network connection. This way we decouple the event store from its users.

### What needs to be done to implement this feature?

- [x] Design an API
  - [x] Clarify if and how the existing domain event API and the event store API can be merged
- [x] Implement the API
  - [x] Add unit tests
- [x] Implement a server that runs the API
  - [x] Add runtime tests

### What else should we know?

Supersedes #100.

PR #399, opened 2019-12-27
feat: Implement domain event store api and microservice process.

yeldiRium:
As discussed in #341 this is an implementation of the query and write APIs for a standalone domain event store service. It consists of the two APIs (and unit tests for them) as well as a process for the microservice runtime (and tests for that).

It is currently a draft, since it relies on [`wait-for-signals`](https://github.com/thenativeweb/wait-for-signals), which we haven't published yet since I just wrote it. Thus the tests will currently fail.


lib/apis/queryDomainEventStore/http/v2/getReplay.ts
```diff
@@ -0,0 +1,74 @@
...
+}: {
+  domainEventStore: DomainEventStore;
+  newDomainEventSubscriber: Subscriber<DomainEvent<DomainEventData>>;
+  newDomainEventSubscriberChannel: string;
+  heartbeatInterval: number;
+}): RequestHandler {
+  return async function (req, res): Promise<any> {
+    let fromRevisionGlobal: number | undefined,
+        observe: boolean,
+        toRevisionGlobal: number | undefined;
```
goloroden: I know that this follows alphabetical order, but IMHO the `from…` and `to…` parameters build a pair and should be closer to each other.

Maybe we should break this up into two blocks?

lib/apis/queryDomainEventStore/http/v2/getReplay.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    try {
+      ({ fromRevisionGlobal, toRevisionGlobal, observe } = parseGetReplayQueryParameters({ parameters: req.query }));
+    } catch (ex) {
+      return res.status(400).send(ex.message);
+    }
+
+    await streamNdjsonMiddleware({ heartbeatInterval })(req, res, (): void => {
+      // No need for a callback for this middleware.
+    });
```
goloroden: I'd suggest to break this up into two lines: The first one to get the middleware, the second one to actually call it. And maybe it would be a good idea to have a hint in the comment that this is usually the `next` function (first I was thinking about why a callback is needed here until I realized that this is the `next` function).

lib/apis/queryDomainEventStore/http/v2/getReplay.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    const domainEventStream = await domainEventStore.getReplay({ fromRevisionGlobal, toRevisionGlobal });
+
+    domainEventStream.on('data', (data): void => {
+      writeLine({ res, data });
+    });
+
+    domainEventStream.on('end', async (): Promise<void> => {
```
goloroden: I'd highly suggest to replace this with a `for await of` loop, since this makes error handling and unsubscribing and all this stuff way easier, and is way better readable.

lib/apis/queryDomainEventStore/http/v2/getReplayForAggregate.ts
```diff
@@ -0,0 +1,85 @@
...
+}: {
+  domainEventStore: DomainEventStore;
+  newDomainEventSubscriber: Subscriber<DomainEvent<DomainEventData>>;
+  newDomainEventSubscriberChannel: string;
+  heartbeatInterval: number;
+}): RequestHandler {
+  return async function (req, res): Promise<any> {
+    let fromRevision: number | undefined,
+        observe: boolean,
+        toRevision: number | undefined;
```
goloroden: Same as above 😉

lib/apis/queryDomainEventStore/http/v2/getReplayForAggregate.ts
```diff
@@ -0,0 +1,85 @@
...
+    try {
+      ({ fromRevision, toRevision, observe } = parseGetReplayForAggregateQueryParameters({ parameters: req.query }));
+    } catch (ex) {
+      return res.status(400).send(ex.message);
+    }
+
+    const { aggregateId } = req.params;
+
+    if (!isUuid(aggregateId)) {
+      return res.status(400).end('Excepted aggregate id to be a uuid.');
```
goloroden: `Excepted` should IMHO be `Expected`.

Or: `Aggregate id must be a UUID.`

lib/apis/queryDomainEventStore/http/v2/getReplayForAggregate.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    await streamNdjsonMiddleware({ heartbeatInterval })(req, res, (): void => {
+      // No need for a callback for this middleware.
```
goloroden: See above 😉

lib/apis/queryDomainEventStore/http/v2/getReplayForAggregate.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    const domainEventStream = await domainEventStore.getReplayForAggregate({ aggregateId, fromRevision, toRevision });
+
+    domainEventStream.on('data', (data): void => {
+      writeLine({ res, data });
+    });
+
+    domainEventStream.on('end', async (): Promise<void> => {
```
goloroden: As above 😉

lib/apis/queryDomainEventStore/http/v2/getReplayForAggregate.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (observe) {
```
goloroden: This should be inverted and then work with a `return` instead of an `else` to improve readability.

lib/apis/queryDomainEventStore/http/v2/parameters/parseGetReplayForAggregateQueryParameters.ts
```diff
@@ -0,0 +1,42 @@
...
+}): {
+    fromRevision: number | undefined;
+    toRevision: number | undefined;
+    observe: boolean;
+  } {
+  if (parameters.fromRevision !== undefined && isNaN(Number(parameters.fromRevision))) {
+    throw new Error(`Expected query parameter 'fromRevision' to be a number.`);
+  }
+  if (parameters.toRevision !== undefined && isNaN(Number(parameters.toRevision))) {
+    throw new Error(`Expected query parameter 'toRevision' to be a number.`);
```
goloroden: I'm not sure how we named exceptions like these elsewhere, but out of a gut feeling, I think I'd prefer something such as `Query parameter … must be …`.

lib/apis/queryDomainEventStore/http/v2/parameters/parseGetReplayForAggregateQueryParameters.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (toRevision !== undefined && toRevision < 1) {
+    throw new Error(`Expected query parameter 'toRevision' to be at least 1.`);
+  }
+  if (fromRevision !== undefined && toRevision !== undefined && toRevision < fromRevision) {
+    throw new Error(`Expected query parameter 'toRevision' to be greater or equal to 'fromRevision'.`);
+  }
+
+  if (parameters.observe !== undefined && ![ 'true', 'false' ].includes(parameters.observe)) {
+    throw new Error(`Expected query parameter 'observe' to be either 'true' or 'false'.`);
+  }
```
goloroden: Would it maybe make sense to do this check with a schema?

lib/apis/queryDomainEventStore/http/v2/parameters/parseGetReplayQueryParameters.ts
```diff
@@ -0,0 +1,42 @@
...
+    throw new Error(`Expected query parameter 'fromRevisionGlobal' to be at least 1.`);
+  }
+  if (toRevisionGlobal !== undefined && toRevisionGlobal < 1) {
+    throw new Error(`Expected query parameter 'toRevisionGlobal' to be at least 1.`);
+  }
+  if (fromRevisionGlobal !== undefined && toRevisionGlobal !== undefined && toRevisionGlobal < fromRevisionGlobal) {
+    throw new Error(`Expected query parameter 'toRevisionGlobal' to be greater or equal to 'fromRevisionGlobal'.`);
+  }
+
+  if (parameters.observe !== undefined && ![ 'true', 'false' ].includes(parameters.observe)) {
```
goloroden: Again, this could be done using a schema.

lib/apis/writeDomainEventStore/http/v2/storeDomainEvents.ts
```diff
@@ -0,0 +1,62 @@
...
+    } catch {
+      return res.status(415).send('Header content-type must be application/json.');
+    }
+
+    if (contentType.type !== 'application/json') {
+      return res.status(415).send('Header content-type must be application/json.');
+    }
+
+    if (!Array.isArray(req.body)) {
+      return res.status(400).send('Excpected array of domain events.');
```
goloroden: To use the same wording as above, this should be renamed to something using `must`.

lib/apis/writeDomainEventStore/http/v2/storeDomainEvents.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    const storedDomainEvents = await domainEventStore.storeDomainEvents({ domainEvents });
+
+    for (const domainEvent of storedDomainEvents) {
+      await newDomainEventPublisher.publish({
+        channel: newDomainEventPublisherChannel,
+        message: domainEvent
+      });
+    }
```
goloroden: We probably talked about this one, but what if the process crashes before publishing, but after storing?

test/runtime/microservice/processes/domainEventStore/processTests.ts
```diff
@@ -0,0 +1,266 @@
...
+      });
+
+      assert.that(postStatus).is.equalTo(200);
+
+      await collector.promise;
+    });
+  });
+
+  suite('GET /query/v2/replay/:aggregateId', (): void => {
+    test('sends out only domain events for the requested aggregate.', async (): Promise<void> => {
```
goloroden: `returns` instead of `sends out`?

test/runtime/microservice/processes/domainEventStore/processTests.ts
```diff
@@ -0,0 +1,266 @@
...
+          assert.that(currentDomainEvent).is.equalTo(domainEvent);
+
+          await collector.signal();
+        }
+      ]));
+
+      await collector.promise;
+    });
+
+    test('sends out all concurrently stored domain events when observing.', async (): Promise<void> => {
```
goloroden: `returns` instead of `sends out`?

test/runtime/microservice/processes/domainEventStore/processTests.ts
```diff
@@ -0,0 +1,266 @@
...
+        method: 'get',
+        url: `http://localhost:${port}/health/v2`
+      });
+
+      assert.that(status).is.equalTo(200);
+    });
+  });
+
+  suite('GET /query/v2/replay', (): void => {
+    test('sends out all previously stored domain events.', async (): Promise<void> => {
```
goloroden: `returns` instead of `sends out`?

test/runtime/microservice/processes/domainEventStore/processTests.ts
```diff
@@ -0,0 +1,266 @@
...
+          await collector.signal();
+        }
+      ]));
+
+      await collector.promise;
+    });
+  });
+
+  suite('GET /query/v2/last-domain-event', (): void => {
+    test('returns the latest stored domain event.', async (): Promise<void> => {
```
goloroden: `last` instead of `latest`?

test/unit/apis/queryDomainEventStore/httpTests.ts
```diff
@@ -0,0 +1,1015 @@
...
+          ]));
+
+          await newDomainEventPublisher.publish({
+            channel: newDomainEventSubscriberChannel,
+            message: firstDomainEvent
+          });
+        });
+      });
+
+      test('closes the stream once the given to-revision-global is reached and does not deliver it.', async (): Promise<void> => {
```
goloroden: This IMHO is misleading. The event that matches the `to-revision-global` should still be delivered, so the edges are *including*, not *excluding*.

lib/apis/queryDomainEventStore/http/v2/parameters/parseGetReplayForAggregateQueryParameters.ts
yeldiRium: I tried but we can't cover number-like strings with a schema. That's why I ended up with manually checking things. Also the `toRevision < fromRevision` check can't be done with a schema.

lib/apis/writeDomainEventStore/http/v2/storeDomainEvents.ts
yeldiRium: Then the clients have to reconnect and request the events starting from the last event they saw.

goloroden: Ah, I remember 👍 

Thanks!

PR #408, opened 2019-12-30
feat: Aeonstore interface implementation

### LINKED ISSUE #401: Implement DomainEventStore interface using domain event store service

In #341 we discussed and implemented a stand alone service for the domain event store exposing an http API. Currently the `Repository` still uses the databases directly via the `DomainEventStore` interface.

To avoid rewriting the `Repository` and to allow using the current setup in the future while being able to place the domain event store service between the `Repository` and the databases, we want to add another implementation of the `DomainEventStore` interface which uses the new APIs instead of talking directly to a database.

- [ ] Create `DomainEventStore` implementation that talks to the domain event store APIs
  - [ ] Add integration test setup (which will require starting the service somehow)

yeldiRium:
Hi @goloroden,

I'm done with the Aeonstore implementation of the `DomainEventStore` interface. Since the integration tests for the domain event store implementations are much more comprehensive than the tests I had written for the domain event store APIs, a few issues arose and had to be fixed in said APIs.

Meanwhile I also added a utility for filtering out heartbeats from `application/x-ndjson` streams and converting those streams into object streams. This should be re-usable for clients later on.

If you're fine with the changes you can go ahead and merge them.

If merged, this closes #401.


lib/apis/queryDomainEventStore/http/v2/getLastDomainEvent.ts
```diff
@@ -7,7 +7,7 @@ const getLastDomainEvent = function ({
 }: {
   domainEventStore: DomainEventStore;
 }): RequestHandler {
-  return async function (req, res): Promise<void> {
+  return async function (req, res): Promise<any> {
```
goloroden: I assume that this change is caused by the inlined `return` when sending a `404`, isn't it? If so, I would prefer the old version, and suggest to not inline the `return` statement. Usually, I'm a big of this, but in this case this has a side-effect I pretty much dislike. What do you think?

lib/apis/writeDomainEventStore/http/v2/storeDomainEvents.ts
```diff
@@ -46,16 +50,20 @@ const storeDomainEvents = function ({
...
+        await newDomainEventPublisher.publish({
+          channel: newDomainEventPublisherChannel,
+          message: domainEvent
+        });
+      }
 
-    res.json(storedDomainEvents);
+      res.json(storedDomainEvents);
+    } catch (ex) {
+      return res.status(400).send(ex.message);
```
goloroden: This return is unneeded, is it?

lib/stores/domainEventStore/Aeonstore/AeonstoreDomainEventStore.ts
```diff
@@ -0,0 +1,137 @@
...
+    const { data } = await axios({
+      method: 'get',
+      url: `${this.aeonstoreBaseUrl}/query/v2/replay/${aggregateId}${queryString}`,
+      responseType: 'stream'
+    });
+
+    const passThrough = new PassThrough({ objectMode: true });
+    const heartbeatFilter = new FilterHeartBeatsFromJsonStreamTransform();
+
+    return data.pipe(heartbeatFilter).pipe(passThrough);
```
goloroden: We should not use `pipe`, since it does not correctly unwind things when something goes wrong. We should use the `pipeline` function of the `stream` module instead.

lib/stores/domainEventStore/Aeonstore/AeonstoreDomainEventStore.ts
```diff
@@ -0,0 +1,137 @@
...
+    const { data } = await axios({
+      method: 'get',
+      url: `${this.aeonstoreBaseUrl}/query/v2/replay${queryString}`,
+      responseType: 'stream'
+    });
+
+    const passThrough = new PassThrough({ objectMode: true });
+    const heartbeatFilter = new FilterHeartBeatsFromJsonStreamTransform();
+
+    return data.pipe(heartbeatFilter).pipe(passThrough);
```
goloroden: Same as above, we should use `stream.pipeline` instead of `pipe`.

lib/apis/queryDomainEventStore/http/v2/getLastDomainEvent.ts
yeldiRium: It actually should have been this way before already but I forgot I return for the 400 status code. We have it like this in almost all api endpoints.

goloroden: Okay 👍

lib/apis/writeDomainEventStore/http/v2/storeDomainEvents.ts
yeldiRium: It is, but i'm in favor of always putting a `return` in front of lines that end the connection so that code added later cannot accidentally continue after these lines.

PR #483, opened 2020-01-31
Fix #131 event ordering

### LINKED ISSUE #131: Events may not get ordered correctly
### What is this bug about?

When storing events in the event store, we write a `position` field to get a global ordering for replaying. Until now we assumed that this is safe, since the `position` is incremented automatically (except in MongoDB), and we assumed that accessing the sequence is safe due to the surrounding transaction. As it turns out, this is *not* the case (see [this StackOverflow question](https://stackoverflow.com/questions/2095917/sequences-not-affected-by-transactions) for details). Unfortunately, this not only affects PostgreSQL, but basically all databases we support.

A simple way to solve this issue would be to acquire a table lock, but this is pretty expensive. As a cheap alternative, we may use `advisory locks`, which exist for all relational databases we support:

- [PostgreSQL](https://www.postgresql.org/docs/9.4/explicit-locking.html#ADVISORY-LOCKS)
- [SQL Server](https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-getapplock-transact-sql?view=sql-server-2017)
- [MySQL](https://dev.mysql.com/doc/refman/8.0/en/locking-functions.html#function_get-lock)
- [MariaDB](https://mariadb.com/kb/en/library/get_lock/)

In MongoDB, there is no such lock concept, but since MongoDB doesn't have auto incrementing fields anyway, we need a different approach here. We currently assume that if we introduced transactions for MongoDB, things would work (also see #65).

### What steps are needed to fix the bug?

- [x] Decide whether to use advisory locks or table locking
- [x] Implement locking for relational databases
- [ ] Define a way how to deal with this in MongoDB
- [ ] Implement locking in MongoDB

### What else should we know?

This bug was brought up by @rkaw92 on Slack, and we should notify him once we have solved this issue (he's also on [Twitter](https://twitter.com/rkaw92)).

damienbenon:
Fixes #131.

Opened as a draft for early discussion.


damienbenon:
Test is expected to fail currently when using a persistent storage.

PR #511, opened 2020-02-11
feat: Rewrite lock stores, implement lock service.

yeldiRium:


lib/common/services/getLockService.ts
```diff
@@ -0,0 +1,15 @@
...
+import { LockStore } from '../../stores/lockStore/LockStore';
+
+const getLockService = function ({ lockStore }: {
+  lockStore: LockStore;
+}): LockService {
+  return {
+    acquireLock: lockStore.acquireLock.bind(lockStore),
+    isLocked: lockStore.isLocked.bind(lockStore),
+    renewLock: lockStore.renewLock.bind(lockStore),
+    releaseLock: lockStore.releaseLock.bind(lockStore)
```
goloroden: Can we make this more explicit, so that we explicitly decouple the store's parameters from the service? This way, if we change the store's interface, it gets more apparent that you are also changing the service interface.

lib/stores/lockStore/InMemory/InMemoryLockStore.ts
```diff
@@ -19,95 +18,79 @@ class InMemoryLockStore implements LockStore {
...
-    namespace: string;
-    value: any;
+    name: string;
     expiresAt?: number;
-    onAcquired? (): void | Promise<void>;
   }): Promise<void> {
-    const name = this.getLockName({ namespace, value });
+    if (name.length >= this.maxLockSize) {
+      throw new errors.LockNameTooLong('Lock name is too long.');
+    }
```
goloroden: I would have expected that the max lock size is still fine to use. So this should be a `>`, not a `>=`, shouldn't it?

lib/stores/lockStore/InMemory/InMemoryLockStore.ts
```diff
@@ -19,95 +18,79 @@ class InMemoryLockStore implements LockStore {
...
-  public async isLocked ({ namespace, value }: {
-    namespace: string;
-    value: any;
+  public async isLocked ({ name }: {
+    name: string;
   }): Promise<boolean> {
-    const name = this.getLockName({ namespace, value });
+    if (name.length >= this.maxLockSize) {
+      throw new errors.LockNameTooLong('Lock name is too long.');
+    }
```
goloroden: As above.

lib/stores/lockStore/InMemory/InMemoryLockStore.ts
```diff
@@ -19,95 +18,79 @@ class InMemoryLockStore implements LockStore {
...
-    namespace: string;
-    value: any;
+  public async renewLock ({ name, expiresAt }: {
+    name: string;
     expiresAt: number;
   }): Promise<void> {
-    const name = this.getLockName({ namespace, value });
+    if (name.length >= this.maxLockSize) {
+      throw new errors.LockNameTooLong('Lock name is too long.');
+    }
```
goloroden: As above.

lib/stores/lockStore/InMemory/InMemoryLockStore.ts
```diff
@@ -19,95 +18,79 @@ class InMemoryLockStore implements LockStore {
...
   }
 
-  public async releaseLock ({ namespace, value }: {
-    namespace: string;
-    value: any;
+  public async releaseLock ({ name }: {
+    name: string;
   }): Promise<void> {
-    const name = this.getLockName({ namespace, value });
+    if (name.length >= this.maxLockSize) {
```
goloroden: As above.

lib/stores/lockStore/MariaDb/MariaDbLockStore.ts
```diff
@@ -116,23 +114,18 @@ class MariaDbLockStore implements LockStore {
...
-    value: any;
+    name: string;
     expiresAt?: number;
-    onAcquired? (): void | Promise<void>;
   }): Promise<void> {
-    const sortedSerializedValue = JSON.stringify(sortKeys({
-      object: value,
-      recursive: true
-    }));
+    if (name.length >= this.maxLockSize) {
```
goloroden: As above.

This occurs even more, but I won't comment this now over and over again 😉

lib/stores/lockStore/MariaDb/MariaDbLockStore.ts
```diff
@@ -164,46 +156,31 @@ class MariaDbLockStore implements LockStore {
...
+        VALUES (?, ?, ?);`;
       } else {
         query = `
         UPDATE ${this.tableNames.locks}
            SET expiresAt = ?,
                nonce = ?
-         WHERE namespace = ?
-           AND value = ?;`;
+         WHERE name = ?;`;
       }
```
goloroden: We have a race condition here, since the lock could have been released since we checked for its existence. Maybe this should be rewritten using an `UPSERT`?

Or – we first check whether there is an expired lock, and if so, we delete it. And then, afterwards, we do an INSERT. This would also remove the problem.

We will have the same problem at other places, too, but I won't comment this every single time.

yeldiRium: Let's do this in a separate PR to not mix this refactor with a fix.

lib/common/services/getLockService.ts
yeldiRium: It uses the `LockService` interface, which is separate from the `LockStore` interface. So developers won't see the `LockStore` interface at all, if they just interface with the `Service`. Isn't that enough? Yes, if one of the services changes, this will have to be rewritten. But as long as the two interfaces are so similar, why make it more complicated here?

goloroden: `bind` is something different than using a new function here which calls the store's method. With `bind` we're essentially passing through the store's method, and redefine `this` (which can be dangerous, as the store's method may not expect this behavior). Hence, I like an explicit boundary more than an implicit one, so I wouldn't call it more complicated, rather more clear 😉 

But I assume this is a question of one's very own flavour. So, to cut a long story short, I'd leave it up to you to decide which you like better. If you want to stick to `bind`, I'm fine, and if you are going to change it, I'm also fine with it.

test/integration/stores/lockStore/getTestsFor.ts
```diff
@@ -62,7 +62,7 @@ const getTestsFor = function ({ createLockStore, inMemory = false, maxLockSize }
     test('acquires a lock with the maximum accepted size.', async (): Promise<void> => {
       // A JSON serialized string will embed opening and closing quotes
       // Those two characters are part of the lock name.
```
goloroden: This comment can be removed, can't it?

test/integration/stores/lockStore/getTestsFor.ts
```diff
@@ -62,7 +62,7 @@ const getTestsFor = function ({ createLockStore, inMemory = false, maxLockSize }
...
-      const maxName = 'a'.repeat(maxLockSize - 2);
+      const maxName = 'a'.repeat(maxLockSize - 1);
```
goloroden: The test title says `with the maximum accepted size`… why the `-1` here?

yeldiRium: Because we changed the condition in the stores back to a `<` instead of a `<=` and thus the maximum is an exclusive maximum.

goloroden: I wasn't sure, so I tested it this morning 😉 

Actually, we do  not check using `<`, but using `>`: We verify if the given lock size is larger than the max allowed one, and if so, throw an error.

So, if you use a lock with the exact size of the max length, it works. I'm going to update this line, and then merge 😊

PR #548, opened 2020-02-23
chore: Add automated version checks

goloroden:


lib/clis/concierge/versions/verify/getBaseImageVersionFromDockerfile.ts
```diff
@@ -0,0 +1,24 @@
+import { errors } from '../../../../common/errors';
+import fs from 'fs';
+
+const getBaseImageVersionFromDockerfile = async function ({ dockerfilePath }: {
+  dockerfilePath: string;
+}): Promise<string> {
+  const dockerfile = await fs.promises.readFile(dockerfilePath, 'utf8');
+
+  const from = dockerfile.split('\n').find((line): boolean => line.startsWith('FROM'));
```
yeldiRium: Should we consider multi-stage builds here and handle multiple `FROM` lines?
If we start managing the versions of all used images, this could be useful.

goloroden: Yes, I'll change that 👍

PR #595, opened 2020-03-06
fix: Support numbers in application names. Closes #585

### LINKED ISSUE #585: Support numbers in application names
### What is this bug about?

Right now, wolkenkit 4 CTP 1 limits application names to only consist of lowercase characters and dashes. However, we should also allow numbers.

### What is the expected result?

The application should be created.

### What is the actual result?

The CLI denies creating the application due to an invalid application name.

### What steps are needed to reproduce the bug?

- Run `npx wolkenkit@4.0.0-ctp.1 init wk4-test`

### What else should we know?

This issue was brought up by @mattwagl, so we should notify him once this bug has been fixed.

goloroden:


goloroden:
@yeldiRium Please don't merge yet!

goloroden:
@yeldiRium This one is good to merge now 😊

PR #600, opened 2020-03-08
Fix command handler authorization.

damienbenon:
* Pass the state value instead of the CurrentAggregateState when calling `command#isAuthorized`.
* Alter the authorize test command to reject authorization when the state property is not found.
* Add test to verify that authorization passes with the correct state value.

This fixes a bug I encountered by adding custom authorization to an `amend` command on the sample `message` aggregate by using typescript.

The `CommandHandler` interface expects a `TState extends State` however `State` is a type redirecting to `object`.

This makes TS a bit confused with this defaulting to `object` and making the compiler accept to pass the `CurrentAggregateState<TState>` instead of the `TState` to the `isAuthorized` handler.

### DIFF
--- a/test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
+++ b/test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
@@ -13,11 +13,18 @@ const authorize = {
   },
 
   isAuthorized (state, command) {
-    return command.shouldAuthorize;
+    // Reject authorization if we cannot see the state properties.
+    // That allows to verify that `handleCommand` passes the state value
+    // instead of the `CurrentAggregateState` wrapper object.
+    if (!Object.prototype.hasOwnProperty.call(state, 'domainEventNames')) {
+      return false;
+    }
+
+    return command.data.shouldAuthorize;
   },
 
   handle (state, command, { aggregate }) {
-    aggregate.publishDomainEvent('authorized');
+    aggregate.publishDomainEvent('authorized', {});
   }
 };
 
--- a/test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
+++ b/test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
@@ -13,10 +13,10 @@ const authorize = {
   },
 
   isAuthorized (state, command) {
-    // Reject authorization if we cannot see the state properties.
-    // That allows to verify that `handleCommand` passes the state value
-    // instead of the `CurrentAggregateState` wrapper object.
-    if (!Object.prototype.hasOwnProperty.call(state, 'domainEventNames')) {
+    // Reject authorization if we cannot see the state properties. That allows
+    // to verify that `handleCommand` passes the state value instead of the
+    // `CurrentAggregateState` wrapper object.
+    if (!state.domainEventNames) {
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
```diff
@@ -13,11 +13,18 @@ const authorize = {
   },
 
   isAuthorized (state, command) {
-    return command.shouldAuthorize;
+    // Reject authorization if we cannot see the state properties.
+    // That allows to verify that `handleCommand` passes the state value
+    // instead of the `CurrentAggregateState` wrapper object.
```
goloroden: Here, to follow our general styleguide, it would be great if you could re-format these lines so that they try to use the most out of 80 characters width. Can you please update the line breaks?

test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
```diff
@@ -13,11 +13,18 @@ const authorize = {
...
+    if (!Object.prototype.hasOwnProperty.call(state, 'domainEventNames')) {
```
goloroden: Although it's not exactly the same, I think I'd prefer the more easily readable version of this without `hasOwnProperty`:

```suggestion
    if (!state.domainEventNames) {
```

test/unit/common/domain/handleCommandTests.ts
```diff
@@ -128,6 +128,61 @@ suite('handleCommand', (): void => {
         await domainEventStore.getLastDomainEvent({ aggregateIdentifier })
       ).is.undefined();
     });
+
+    test(`passes the correct state to the authorize handler.`, async (): Promise<void> => {
```
goloroden: This should be a single-quoted string, as we don't have any quotes inside of it.

```suggestion
    test('passes the correct state to the authorize handler.', async (): Promise<void> => {
```

test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
```diff
@@ -13,11 +13,18 @@ const authorize = {
...
+      return false;
+    }
+
+    return command.data.shouldAuthorize;
   },
 
   handle (state, command, { aggregate }) {
-    aggregate.publishDomainEvent('authorized');
+    aggregate.publishDomainEvent('authorized', {});
```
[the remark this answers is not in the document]
damienbenon: `AggregateService#publishDomainEvent` will complain about `reason: 'Invalid type: undefined should be object (at data).'` as the schema of the event `authorized` declares that it expects an object but nothing is provided. 

Please see `lib/common/services/getAggregateService.ts#47` and schema definition in `test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/domainEvents/authorized.js`).

Actually this could not be detected without having a test that passes the authorization to verify that...

test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
damienbenon: This should be fixed by now.

test/shared/applications/javascript/base/build/domain/sampleContext/sampleAggregate/commands/authorize.js
damienbenon: This should be fixed by now.

test/unit/common/domain/handleCommandTests.ts
damienbenon: This should be fixed by now.

PR #738, opened 2020-04-29
feat: Add create-deployment cli command.

### LINKED ISSUE #546: Add a CLI command to (re)create the deployment scripts

Sometimes, people delete the `deployment` folder, or they don't even have it, e.g. because they migrated an old application from wolkenkit 3.x. For these scenarios, it would be helpful if there was an easy way to (re)create the deployment scripts using the CLI.

- [x] Add a new `create-deployment` command
  - [x] Deny to work if the `deployment` directory already exists
  - [x] Let the user override the existing directory with the `--force` flag
- [x] Add tests

### What else should we know?

Supposed the user provides the `--force` flag, we still need to clarify what to do with existing files. Overwrite them, or backup them, or just ignore them?

yeldiRium:
This closes #546.


lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,105 @@
...
+import { getAbsolutePath } from '../../../common/utils/path/getAbsolutePath';
+import { nameRegularExpression } from '../init/nameRegularExpression';
+import path from 'path';
+import shell from 'shelljs';
+import { validateName } from '../init/validateName';
+
+const createDeploymentCommand = function (): Command<CreateDeploymentOptions> {
+  return {
+    name: 'init',
+    description: 'Initialize a new application.',
```
goloroden, line 17: I assume this is a left-over from copy/paste and needs to be adjusted.

```suggestion
    name: 'create-deployment',
    description: 'Create deployment manifests.',
```

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,105 @@
...
+        parameterName: 'path',
+        type: 'string',
+        isRequired: false
+      }, {
+        name: 'deployment-directory',
+        description: 'the name of the directory in which to create the deployment manifests',
+        type: 'string',
+        isRequired: false,
+        defaultValue: 'deployment'
+      }, {
```
goloroden, line 33: Hmmm 🤔 … to be honest, I found this a little bit hard to understand: Why do I need to provide two directories? Respectively, what happens if I provide absolute paths for both? And similar questions…

What do you think of this suggestion: Instead of two options, we merge this into a single option, `deployment-directory` (aliased as `d`). If the user does *not* provide this option, the default is `./deployment`. If the user provides the option, we take this path (be it a relative or an absolute one).

What do you think of this?

PS: If the user runs this command from a directory which does *not* contain a wolkenkit application, we could either reject this, or we could require that the `--name` flag is given as well. Since in the future, we may rely on more things here, I'd suggest to simply reject it. Of course this then means that you *have* to be in a directory which contains an application, but I assume (yes, this is another guess…) that typically you will want to do this for an application you are currently working on. At least you will always want to do this for an *existing* application.

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,105 @@
...
+        defaultValue: false
+      }, {
+        name: 'name',
+        alias: 'n',
+        description: 'set an application name',
+        type: 'string',
+        isRequired: false,
+        defaultOption: true,
+        validate: validateName
+      }
```
goloroden, line 48: See notes from above.

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,105 @@
...
+          withVerboseMode(verbose)
+      );
+      const stopWaiting = buntstift.wait();
+
+      try {
+        let selectedName = name;
+
+        if (!selectedName) {
+          selectedName = await buntstift.ask('Enter the application name:', nameRegularExpression);
+        }
```
goloroden, line 69: If we only allow this from an application directory (see above), this wouldn't be needed…

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        const targetDirectory = path.join(appDirectory, deploymentDirectory);
+
+        const targetDirectoryExists = await exists({ path: targetDirectory });
+
+        if (targetDirectoryExists && !force) {
+          buntstift.info(`Could not create deployment manifests since the deployment directory ${targetDirectory} already exists.`);
+          buntstift.info('Use the --force flag to overwrite the directory.');
+
+          return process.exit(1);
```
goloroden, line 84: I think in this case the spinner doesn't get stopped. We should add a `stopWaiting()` call here.

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        if (targetDirectoryExists) {
+          shell.rm('rf', targetDirectory);
+        }
```
goloroden, line 89: This comes unexpected – I would have expected that existing files get overwritten, but other files are kept. This deletes the entire content of the deployment directory, including any user-generated files. As said, it's not wrong per se – it's just unexpected.

While thinking about this I was wondering if we should actually allow to overwrite things. If you just one something additional, you could use a new directory, and if you would like to re-create the original one from scratch, you could delete it manually before.

What do you think?

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
yeldiRium: > Since in the future, we may rely on more things here, I'd suggest to simply reject it.

That is probably the crucial point here. I went a bit back and forth on whether to force the user to run the command in the application directory and decided, not to force that, since init and other commands don't do it as well. But that leads to either requiring multiple directory paths/names or requiring additional info via parameters that would be present in the application.

Both options are tedious and developers will most likely run commands close to where they need them anyway. So I'll adjust the PR like you suggested - rejecting running the command outside of application directories. This will remove the `name` and `directory` parameters and only leave the `deployment-directory` parameter, which can then be an actual path instead of just a directory name.

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
yeldiRium: Yes, I missed those.

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
yeldiRium: > I would have expected that existing files get overwritten, but other files are kept

Good point. I did not think about the possibility, that developers add new files to the directory. You are probably right, that we should not delete/overwrite stuff and instead reject and ask for another directory name.

test/cli/createDeploymentTests.ts
```diff
@@ -0,0 +1,84 @@
+import { assert } from 'assertthat';
+import { exists } from '../../lib/common/utils/fs/exists';
+import { isolated } from 'isolated';
+import path from 'path';
+import shell from 'shelljs';
+
+const rootPath = path.join(__dirname, '..', '..');
+const cliPath = path.join(rootPath, 'build', 'lib', 'bin', 'wolkenkit.js');
+
+suite('init', function (): void {
```
goloroden, line 10:
```suggestion
suite('create-deployment', function (): void {
```

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const createDeploymentCommand = function (): Command<CreateDeploymentOptions> {
+  return {
+    name: 'create-deployment',
+    description: 'Create deployment manifests for a wolkenkit application.',
+
+    optionDefinitions: [
+      {
+        name: 'deployment-directory',
+        description: 'the directory in which to create the deployment manifests',
```
goloroden, line 19: To streamline this message with those of the other commands:

```suggestion
        description: 'set a deployment directory',
```

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,73 @@
...
+      verbose,
+      'deployment-directory': deploymentDirectory
+    }}): Promise<void> {
+      buntstift.configure(
+        buntstift.getConfiguration().
+          withVerboseMode(verbose)
+      );
+
+      try {
+        const appDirectory = process.cwd();
```
goloroden, line 37:
```suggestion
        const applicationDirectory = process.cwd();
```

This needs to be changed in some additional places as well…

lib/clis/wolkenkit/createDeployment/createDeploymentCommand.ts
```diff
@@ -0,0 +1,73 @@
...
+        try {
+          const packageJsonPath = path.join(appDirectory, 'package.json');
+          const packageJsonContent = JSON.parse(await fs.promises.readFile(packageJsonPath, 'utf-8'));
+
+          applicationName = packageJsonContent.name;
+        } catch {
+          buntstift.info('Failed to create deployment manifests.');
+          buntstift.info('Please run the create-deployment command in a wolkenkit application directory.');
+
+          return process.exit(1);
```
goloroden, line 49: I'm not quite sure on the verbosity of these messages here, but I think let's do it like this, and I'll probably go through this at a later point in time again. Just wanted to let you know already 😊

PR #827, opened 2020-05-25
Include sourcemaps for Typescript users.

### LINKED ISSUE #783: Enable source map support for TypeScript applications

While debugging got a lot easier in 4.0, debugging TypeScript still does not feel as good as it could. To improve the developer experience here, we should enable source map support for TypeScript applications. This probably comes down to just adding the appropriate setting to the `tsconfig.json` file:

```json
{
  "compilerOptions": {
    "sourceMap": true
  },
  "include": [
    "./server/**/*.ts"
  ]
}
```

- [x] Figure out how to enable source map support for TypeScript

### What else should we know?

This issue was brought up by @damienbenon, so we should notify him once this has been done.

damienbenon:
Fixes #783.

It includes now `sourcemap` files along with typescript definitions.

This allows users to step through Wolkenkit sources during debug.

The published Wolkenkit package already embeds both typescript sources (`lib`) and compiled output (`build/lib`) so no additional work than publishing a new version should be required.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,7 @@
     "module": "CommonJS",
     "outDir": "build",
     "resolveJsonModule": false,
+    "sourceMap": true,
     "strict": true,
     "target": "ES2019"
   },
--- a/templates/blank/typescript/tsconfig.json
+++ b/templates/blank/typescript/tsconfig.json
@@ -1,4 +1,7 @@
 {
+  "compilerOptions": {
+    "sourceMap": true
+  },
   "include": [
     "./server/**/*.ts"
   ]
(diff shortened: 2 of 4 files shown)


tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "outDir": "build",
     "resolveJsonModule": false,
     "strict": true,
-    "target": "ES2019"
+    "target": "ES2019",
+    "sourceMap": true
```
goloroden, line 12: Can you please sort this in alphabetical order?

damienbenon: Sure, this is now updated.

damienbenon:
Hi @goloroden,

Sorry, I just missed the usual conventions you use on Wolkenkit!

I've also added on your suggestion the sourcemap option on application templates.

PR #907, opened 2020-06-15
feat: Split handleCommand api into separate routes for commands.

### LINKED ISSUE #565: Introduce separate routes for the individual commands

For 4.0, so far we have designed the command API to have just a single `/command` endpoint. This has historical reasons. Since we dropped the entire command metadata on the client-side, we could now go with an individual route per command, such as

```
POST /:contextName/:aggregateName/:aggregateId/:commandName
```

and transfer the command's data in the `POST` body.

- [x] Change the `handleCommand` server so that it provides multiple routes instead of a single one
- [x] Ensure to correctly build the `CommandWithMetadata` instances
- [x] Adjust tests

yeldiRium:
Closes #565.

### DIFF
--- a/lib/apis/handleCommand/http/v2/postCommand.ts
+++ b/lib/apis/handleCommand/http/v2/postCommand.ts
@@ -4,7 +4,6 @@ import { Command } from '../../../../common/elements/Command';
 import { CommandWithMetadata } from '../../../../common/elements/CommandWithMetadata';
 import { errors } from '../../../../common/errors';
 import { flaschenpost } from 'flaschenpost';
-import { getCommandSchema } from '../../../../common/schemas/getCommandSchema';
 import { OnReceiveCommand } from '../../OnReceiveCommand';
 import typer from 'content-type';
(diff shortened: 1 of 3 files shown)


lib/apis/handleCommand/http/v2/postCommand.ts
```diff
@@ -16,10 +15,10 @@ const logger = flaschenpost.getLogger();
 
 const postCommand = {
   description: 'Accepts a command for further processing.',
-  path: '',
+  path: ':contextName/:aggregateName/:aggregateId/:commandName',
 
   request: {
-    body: getCommandSchema()
+    body: { type: 'object' }
```
goloroden, line 21: Don't we lose some verification here, e.g. that the `aggregateId` is actually a UUID?

yeldiRium: True. We have no way to validate parameters so far. I'll add a validation against the command schema to the handler.

PR #1107, opened 2020-07-21
feat: Refactor GraphQL; Add query for views.

### LINKED ISSUE #1039: Implement views

Once flows have been implemented, we should next address views.

- [x] Define and update view definitions
- [x] Implement loading and validating views
- [x] Update test applications
- [x] Update sample applications
- [x] Implement views
  - [x] Implement API
  - [x] Add API to microservice runtime
  - [x] Add API to single-process runtime
- [x] Adjust Docker-Compose files
- [x] Add sandbox for views / query handlers
- [x] Add view / query handler tests to sample applications
- [x] Add tests
  - [x] API tests
  - [x] `executeQueryHandler` tests
  - [x] Runtime tests
- [x] Add OpenAPI for views
- [x] Add GraphQL for views
- [x] Add HTTP and GraphQL samples to the `README.md` file

yeldiRium:
This closes #1039, since it resolves the remaining open points.

I've refactored the graphql API to no longer programatically build a schema and the resolvers in parallel. It now uses the graphql-js building blocks to define the schema and the resolvers together, which makes them much more readable and maintainable, as well as more encapsulated.

In addition to that, the new views can now be queried via GraphQL.


README.md
```diff
@@ -189,6 +189,28 @@ subscription {
...
+    all {
+      contextIdentifier {
+        name
+      }
+      aggregateIdentifier {
+        id
+        name
+      }
+      name
+      id
```
goloroden, line 208: Just to get the understanding right – this is specific to the `sampleView`, right?

In the previous examples we used the chat application, shouldn't we do this here as well, and query the `messages/all` view?

yeldiRium: Yes it is and good point. I'll change that.

PR #1128, opened 2020-07-23
feat: Health check command & traefik proxy.

yeldiRium:
Our health check in the docker-compose manifests were broken, since `curl` was not available in any of the containers. To remove this dependency on external tools, I've added the `health` command to the `wolkenkit` cli. Its default options fit the `wolkenkit dev` options. Other than that, the protocol, host name, port and path can be overridden via cli arguments.

This health check command is now used in the docker-compose manifests.

I've also added traefik as the reverse proxy for the microservice runtime. This PR is still a draft, because this decision is not final and the configuration might need improvement. But so far it works.


lib/clis/wolkenkit/createDeployment/dockerCompose/getMicroservicePostgresManifest.ts
```diff
@@ -401,15 +400,21 @@ ${
 }
         image: '${appName}'
         init: true
-        ports:
-          - '${services.command.publicPort}:${services.command.privatePort}'
+        expose:
+          - '${services.command.privatePort}'
```
goloroden, line 404: I don't think we need to expose the private ports, especially given that we don't do it for the containers that are accessible only internally, anyway. So we should either remove it here (and at the other places), or introduce it everywhere, for consistency reasons.

Is there a benefit of providing them?

lib/clis/wolkenkit/createDeployment/dockerCompose/getMicroservicePostgresManifest.ts
```diff
@@ -659,6 +688,20 @@ ${
...
+        command:
+          - '--log.level=DEBUG'
+          - '--api.insecure=true'
+          - '--providers.docker=true'
+          - '--providers.docker.exposedbydefault=false'
+          - '--entrypoints.web.address=:3000'
+        ports:
+          - '3000:3000'
+          - '8080:8080'
+        volumes:
```
goloroden, line 703: Traefik is missing a healthcheck for itself, isn't it?

lib/clis/wolkenkit/createDeployment/dockerCompose/getMicroservicePostgresManifest.ts
```diff
@@ -659,6 +688,20 @@ ${
         restart: 'always'
         volumes:
           - 'minio:/data'
+          
+      ${services.traefik.hostName}:
+        image: 'traefik:v2.2'
```
goloroden, line 693: I think that we should set the version number in `versions.ts`, as we do for the other images as well.

lib/clis/wolkenkit/health/healthCommand.ts
```diff
@@ -0,0 +1,88 @@
+import { buntstift } from 'buntstift';
+import { Client } from '../../../apis/getHealth/http/v2/Client';
+import { Command } from 'command-line-interface';
+import { HealthOptions } from './HealthOptions';
+import { validatePort } from '../dev/validatePort';
+
+const healthCommand = function (): Command<HealthOptions> {
+  return {
+    name: 'health',
+    description: 'Run a health check against a process.',
```
goloroden, line 10:
```suggestion
    description: 'Verify the health of a wolkenkit application process.',
```

lib/clis/wolkenkit/health/healthCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    optionDefinitions: [
+      {
+        name: 'protocol',
```
goloroden, line 14: Shouldn't we have aliases here, as we do for the other commands as well?

I'd suggest:

- `r` for `protocol` (to keep `p` for port, which is probably more often needed)
- `n` for `host-name` (we can't use `h` because this is reserved for `help`)
- `p` for `port`
- `b` for `base-path`

lib/clis/wolkenkit/health/healthCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+        description: 'set a port for the health endpoint',
+        parameterName: 'port',
+        type: 'number',
+        isRequired: false,
+        defaultValue: 3001,
+        validate: validatePort
+      },
+      {
+        name: 'base-path',
+        description: 'set the base path of the health api',
```
goloroden, line 41:
```suggestion
        description: 'set the health base path',
```

(To be consistent with the other descriptions)

lib/clis/wolkenkit/health/healthCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+        description: 'set the host name of the server',
+        parameterName: 'hostName',
+        type: 'string',
+        isRequired: false,
+        defaultValue: 'localhost'
+      },
+      {
+        name: 'health-port',
+        alias: 'e',
+        description: 'set a port for the health endpoint',
```
goloroden, line 32:
```suggestion
        description: 'set the health port',
```

lib/clis/wolkenkit/health/healthCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+        name: 'protocol',
+        description: 'set the protocol',
+        parameterName: 'protocol',
+        type: 'string',
+        isRequired: false,
+        defaultValue: 'http'
+      },
+      {
+        name: 'host-name',
+        description: 'set the host name of the server',
```
goloroden, line 23:
```suggestion
        description: 'set the host name',
```

(If we don't specify "of the server"  for the protocol, we don't need it here as well … and I assume it's obvious whose host  name and protocol are needed)

lib/clis/wolkenkit/health/healthCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+      'health-port': healthPort,
+      'base-path': basePath,
+      verbose
+    }}): Promise<void> {
+      buntstift.configure(
+        buntstift.getConfiguration().
+          withVerboseMode(verbose)
+      );
+
+      buntstift.verbose(`Sending health request to ${protocol}://${hostName}:${healthPort}${basePath}.`);
```
goloroden, line 61: Is there a special reason why all messages are typed as `verbose`?

If not, I'd suggest `info` here.
Oh, and the url should be in single quotes.

lib/clis/wolkenkit/health/healthCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+        protocol,
+        hostName,
+        port: healthPort,
+        path: basePath
+      });
+
+      try {
+        const healthData = await healthClient.getHealth();
+
+        buntstift.verbose('Health check successful.');
```
goloroden, line 73: Same as above, should be `success`, unless there is a special reason to keep it as `verbose` level.

lib/clis/wolkenkit/health/healthCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+        buntstift.verbose(JSON.stringify(healthData, null, 2));
+
+        // eslint-disable-next-line unicorn/no-process-exit
+        process.exit(0);
+      } catch (ex) {
+        buntstift.verbose('Health check failed.');
```
goloroden, line 79: Same as above, should be `error`, unless there is a special reason to keep it as `verbose` level.

goloroden:
Hey @yeldiRium 👋 

Thanks for the PR! I've had a look at it and left a number of comments. Can you please review them?

lib/clis/wolkenkit/createDeployment/dockerCompose/getMicroservicePostgresManifest.ts
yeldiRium: We need to expose the ports so that traefik knows about them and can configure its services. Alternatively we can explicitly set these ports in the traefik labels.

lib/clis/wolkenkit/health/healthCommand.ts
yeldiRium: I wanted to only output anything in verbose mode, since the main purpose of the health check for now is scripting. If the command is more intended for users, a `quiet` flag and the opposite behavior would be better.

lib/clis/wolkenkit/createDeployment/dockerCompose/getMicroservicePostgresManifest.ts
goloroden: Ah okay, I wasn’t aware of that. Then let’s leave it like this.

lib/clis/wolkenkit/health/healthCommand.ts
goloroden: For consistency reasons I’d vote for the option you described

yeldiRium: Although in scripting some output doesn't hurt. So we can just make it user-friendly.

lib/clis/wolkenkit/createDeployment/dockerCompose/getMicroservicePostgresManifest.ts
yeldiRium: theoretically yes, but i'm not sure how to do that. in the traefik container neither our healthcheck nor `curl` are available.

goloroden: Hm, okay … and now introducing a custom Traefik container which has `curl` is definitely not the way to go. Okay, then let's leave it like this for now.

yeldiRium: Traefik has a [healthcheck command built into its cli](https://docs.traefik.io/operations/cli/#healthcheck). I'm using that now and it works fine.

PR #1503, opened 2020-10-30
chore: Update dependencies and clean up eslint.

yeldiRium:


lib/common/utils/path/getAbsolutePath.ts
```diff
@@ -1,4 +1,5 @@
 import untildify from 'untildify';
+// eslint-disable-next-line unicorn/import-style
```
goloroden, line 2: Is there a special reason why it is preferable here to disable the linter instead of fixing the problem?

lib/messaging/pubSub/Publisher.ts
```diff
@@ -1,6 +1,7 @@
+// eslint-disable-next-line @typescript-eslint/ban-types
```
goloroden, line 1: This disabling isn't needed any more, is it? We should probably search for `ban-types` and check whether there are more than that where it can be removed.

lib/stores/domainEventStore/MongoDb/MongoDbDomainEventStore.ts
```diff
@@ -150,12 +153,14 @@ class MongoDbDomainEventStore implements DomainEventStore {
     return domainEventCount !== null;
   }
 
-  public async getDomainEventsByCorrelationId <TDomainEventData extends DomainEventData> ({ correlationId }: {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
goloroden, line 156: Why is this needed?

goloroden (changes requested):
Hey @yeldiRium @grundhoeferj 👋 

Thanks for the PR, I've reviewed it (quickly), and found a few things, where I at least have questions. Can you maybe have a second look?

lib/common/utils/path/getAbsolutePath.ts
yeldiRium: yes: the function `getAbsolutePath` below takes a parameter called `path`, so wa can't use that name to import the `path` module here.

lib/stores/domainEventStore/MongoDb/MongoDbDomainEventStore.ts
yeldiRium: Iirc it should not be needed anymore. I'll double check.

lib/messaging/pubSub/Publisher.ts
yeldiRium: Agreed.

lib/common/utils/path/getAbsolutePath.ts
goloroden: Okay 👍

lib/stores/domainEventStore/MongoDb/MongoDbDomainEventStore.ts
goloroden: Thanks 😊

lib/messaging/pubSub/Publisher.ts
goloroden: I fixed this 😊

goloroden (approved):

PR #1821, opened 2021-01-14
chore: Remove create deployment sub command; Replace with templating.

### LINKED ISSUE #1811: Remove create-deployment CLI sub-command; Add templating for deployment
### What is this task about?

We have noticed in and around issue #1805 that the sample applications in their current state partly don't work, since the custom infrastructure (i.e. MongoDB) is not available in the docker-compose manifests generated by the `create-deployment` sub-command. We also cannot create a manifests including MongoDB in these manifests, since the usage of MongoDB is specific to the chat applications and should not appear in the blank sample application.

To work around this, we have decided to remove the CLI subcommand `create-deployment`. We will still generate the manifests that are required by all sample applications and which are currently documented in the readme when initializing a sample application. But we will replace the current mechanism with templating using `ejs`. The templates will be located in the sample applications and can includes shared partials.

This way, additional manifests can be added and manifests can be modified per sample application.

### What needs to be done to complete this task?

- [x] Remove the `create-deployment` sub-command from the `wolkenkit` CLI
- [x] Add the dependency ejs
- [x] Replace the manifest generation logic with templates
- [x] Add the required manifests to the sample applications to set up the MongoDB. This manifests should be called `infrastructure.yml`, to re-use the term and make the connection to the `infrastructure` in a wolkenkit application clear.
- [x] Adjust tests as necessary

yeldiRium:
This closes #1811. It removes the `create-deployment` sub-command on the wolkenkit CLI. Instead, the sample applications that can be used to `init` a new wolkenkit application, now contain docker-compose manifests which use `ejs` for templating.


lib/clis/wolkenkit/init/initCommand.ts
```diff
@@ -133,9 +136,28 @@ const initCommand = function (): Command<InitOptions> {
...
+            await fs.promises.writeFile(
+              path.join(targetDirectory, relativeFilePath.replace('.ejs', '')),
+              renderedFile,
+              'utf-8'
+            );
+          } else {
+            cp(path.join(sourceDirectory, relativeFilePath), path.join(targetDirectory, relativeFilePath));
+          }
+        }
         buntstift.verbose('Copied files.');
```
goloroden, line 161: Shouldn't this be `Copied and rendered files.`?

lib/clis/wolkenkit/init/initCommand.ts
```diff
@@ -133,9 +136,28 @@ const initCommand = function (): Command<InitOptions> {
...
+              path.join(sourceDirectory, relativeFilePath),
+              getTemplateData({ appName: selectedName })
+            );
+
+            await fs.promises.writeFile(
+              path.join(targetDirectory, relativeFilePath.replace('.ejs', '')),
+              renderedFile,
+              'utf-8'
+            );
+          } else {
```
goloroden, line 157: What about inverting the `if` here, and using it with an early `continue` to avoid the nesting of this long block?

goloroden (changes requested):
LGTM 👍 

I've only got two super-small questions, can you maybe have a look at them?

goloroden (approved):

PR #1901, opened 2021-02-15
Environment variable for heartbeat interval

kusigit:
The heartbeat interval should be configurable via an environment variable. E.g. the request timeout on the Heroku Cloud is 55 seconds.

https://devcenter.heroku.com/articles/request-timeout#long-polling-and-streaming-responses

The implementation is just for the single process. Maybe we have to adapt it to microservices as well.


goloroden:
Hey @kusigit 👋 

Thanks a lot for your PR!
We will have a closer look at it next week, and then get back to you.

Have a nice weekend,


Golo

goloroden:
@kusigit So, as promised, we get back to you this week 😊

The PR looks pretty good, however, we have two questions on it:

1. Woud you like to add yourself as a contributor to the `package.json` file?
2. Would you mind adding the same settings to the micro-service runtime?

Cheers,


Golo

lib/runtimes/microservice/processes/notification/configurationDefinition.ts
```diff
@@ -26,6 +26,11 @@ const configurationDefinition: ConfigurationDefinition<Configuration> = {
     defaultValue: 3_001,
     schema: portOrSocketSchema
   },
+  heartbeatInterval: {
+    environmentVariable: 'HEARTBEAT_INTERVAL',
+    defaultValue: 90_000,
+    schema: { type: 'integer' }
```
goloroden, line 32: Should we maybe extract this to a `heartbeatIntervalSchema`, similar to how we did it with ports?

And: Shouldn't we set a `minimum` in that schema, then?

(The same applies to the other places where this parameter is defined.)

goloroden (changes requested):
Hey @kusigit 👋 

Thanks for the quick update, this looks pretty good IMHO 😊 

There is one last thing (see my comment). Would you mind to have another look?

goloroden (approved):

goloroden:
@kusigit Thanks for the quick update 😊

@yeldiRium To me, this PR looks fine – can you please have another look at it, and (if you are fine with it), merge it?

lib/runtimes/microservice/processes/notification/configurationDefinition.ts
kusigit: Hi @goloroden 
You're right. Good idea. Thanks.

yeldiRium (approved):

yeldiRium:
LGTM :+1:

PR #1933, opened 2021-03-03
Default value for ssl property (Postgres)

kusigit:
The default value for the ssl property should not be set. Some cloud providers use TLS but without verification. 
https://devcenter.heroku.com/articles/heroku-postgresql#connecting-in-node-js

I testet my change without TLS (localhost) and with TLS (Heroku) connection.

PostgreSQL 13.2 (without TLS)
PostgreSQL 13.1 (TLSv1.2)


goloroden:
Oh, okay, actually I think we weren't aware that it makes a difference whether you set it to `false` or leave it `undefined`. Why is this?

kusigit:
Hi @goloroden 

I'm not 100% sure, but it looks like if the ssl property is set to _undefined_, some TLS properties from container/runtime engine will be used. 


`ssl = false`

false

`ssl = undefined` (not set)

undefined

goloroden (approved):

goloroden:
@kusigit Makes sense, so I am going to merge this one. Thanks for figuring this out 😊

Issue #1898, opened 2021-02-11
Ensure that the priority queue runs stable


goloroden:
### What is this task about?

From time to time, we have the problem that the priority queue does not behave stable. This sometimes makes the tests fail, but we do not yet know why. We have to figure out what the problem here is.

### What needs to be done to complete this task?

- [ ] Figure out whether this affects all implementations, or just some of them
- [ ] Find the root cause of the error (this probably should be done together with @strangedev and @dotKuro)
- [ ] Fix it

strangedev:
We built some fuzzing tests for our various priority queue implementations and simulated load for about 2 hours on each implementation (#1935). We didn't find anything related to this issue, so we're closing this for the time being, until we get further reports which we can investigate.

PR #1935, opened 2021-03-04
chore: add a fuzzer for the priority queue

strangedev:
This is related to #1898, where we had noticed that the priority queue crashes in tests sometimes. We weren't sure what caused these crashes and had no logs, so we decided to fuzz-test our various priority queue store implementations.

This PR adds the fuzz-test for all store implementations.


test/fuzzing/priorityQueueStore/naughtyStrings.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'big-list-of-naughty-strings' {
+  const nuaghtyStrings: string[];
+  export default nuaghtyStrings;
```
goloroden, line 3: Typos 😉 

This should be `naughty`, not `nuaghty`, I guess…

Besides that – why do we need this?

lib/stores/priorityQueueStore/Observer/PriorityQueueObserver.ts
```diff
@@ -0,0 +1,275 @@
...
+
+interface Issue {
+  type: 'action' | 'issue' | 'error';
+  message: string;
+  data?: any;
+}
+
+const errors = defekt({
+  ObserverError: {}
+});
```
goloroden, line 16: Why do we have the separation here between errors and wolkenkit errors?

goloroden (changes requested):

test/fuzzing/priorityQueueStore/naughtyStrings.d.ts
strangedev: We tried to use https://www.npmjs.com/package/big-list-of-naughty-strings through npm first, but decided against it, because the package doesn't seem well-maintained. This definitions file is a leftover from this approach.

goloroden: Thanks for clarifying this 😊

lib/stores/priorityQueueStore/Observer/PriorityQueueObserver.ts
strangedev: We didn't want to include an error that can never happen in production in the other errors. We use `defekt` here to include some information with the error and have better error messages when the fuzzer finds a problem. I guess we could also include this with the other errors, if you think that's a better design.

goloroden: What if we turn this around? If we leave `wolkenkitErrors` as `errors`, so that it is consistent with everywhere else? And rename this `errors` to something that describes the special situation we're in more descriptive?

goloroden (approved):

PR #2196, opened 2021-06-16
Postgres SSL property

kusigit:
The ssl property of the node postgres driver can be undefined (e.g. localhost) or an object.

https://node-postgres.com/features/ssl

https://help.heroku.com/MDM23G46/why-am-i-getting-an-error-when-i-upgrade-to-pg-8

image


yeldiRium:
Hi @kusigit,

thanks for this PR. Sorry for the delay. I finally had some time to take a look at it and have some questions and annotations.

- You extended the JSON schemas to validate certain fields on the `encryptConnection` object. According to the schema, it now has to be an object. According to the `pg` TypeScript types, the `encryptConnection` option is a `boolean | ConnectionOptions`. Is there a reason that you excluded the possible `boolean` type?
- In the schema you listed four properties, `rejectUnauthorized`, `ca`, `key` and `cert`. Is there a reason why you chose these four specifically? The `ConnectionOptions` type has many more properties that we could also allow here, but I don't know what most of them do.
- If we're going to support certain options, we'll need a type for them. Just using `object` as the type for the `encryptConnection` option feels a bit unsafe and we have no TypeScript support when passing these options to the Postgres constructor.
- I also had a look at your PR in `command-line-interface` (https://github.com/thenativeweb/command-line-interface/pull/365) and unfortunately doubt that this approach will work. So far we pass objects to environment variables or CLI options in the form of JSON strings, because I don't see a way to pass an object to a CLI. We should keep this approach and use `string` Options instead, which are then parsed and validated. (See e.g. the `replay` command.) This would make the PR in `command-line-interface` obsolete.

Do you mind elaborating on my questions? Then we can see what needs to be changed in the PR :)

Thanks again and have a nice day

kusigit:
Hello @yeldiRium

Thanks for the review. You are absolutely right about your comments. I changed the type for the `encryptConnection`. In most cases the four properties should be enough. Let me know if you think we should support them all.

Greetings

yeldiRium:
Thanks for the update, @kusigit :)

I made some changes as well.

- I adjusted the CLI options a bit. The interface for the options must only contain the primitive base types the CLI framework can handle, so we can't use the `ConnectionOptions` interface there. We have to parse and validate the incoming string.
- I also added some tests that the option is parsed correctly and that connecting to a database via SSL works.

I dislike putting certificate files into a repository, but in this case it's only self-signed certs for the tests and their validity is ignored, so it should be fine.

@goloroden do you want to do a final review?

goloroden:
@kusigit @yeldiRium Sorry for the very late reply – I've cleaned up lots of PRs this evening, and I'll come to this tomorrow 😊

lib/stores/consumerProgressStore/Postgres/PostgresConsumerProgressStoreOptions.ts
```diff
@@ -7,6 +8,6 @@ export interface PostgresConsumerProgressStoreOptions {
   userName: string;
   password: string;
   database: string;
-  encryptConnection?: boolean;
+  encryptConnection?: boolean | ConnectionOptions;
```
goloroden, line 11: Is this actually `boolean | …`, or shouldn't it rather be `false | …`?

lib/common/schemas/getConnectionOptionsSchema.ts
```diff
@@ -0,0 +1,23 @@
...
+const getConnectionOptionsSchema = function (): Schema {
+  return {
+    oneOf: [
+      {
+        type: 'object',
+        properties: {
+          rejectUnauthorized: { type: 'boolean' },
+          ca: { type: 'string' },
+          key: { type: 'string' },
+          cert: { type: 'string' }
```
goloroden, line 12: Should we maybe rename this to `privateKey` and `certificate`, to make it more readable?

lib/common/schemas/getConnectionOptionsSchema.ts
```diff
@@ -0,0 +1,23 @@
...
+        },
+        additionalProperties: false
```
goloroden, line 14: The `required` field is missing here (and I think we should even add it if nothing is mandatory, just to point out that we did not forget to add `required` 😉 ).

docker/wolkenkit-postgres-ssl/server.req
```diff
@@ -0,0 +1,60 @@
+Certificate Request:
```
goloroden, line 1: I don't think that we need to keep the CSR, do we?

docker/wolkenkit-postgres-ssl/server.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
goloroden, line 1: To be in line with other places where we use certificates, this should rather be `privateKey.pem`.

docker/wolkenkit-postgres-ssl/server.crt
```diff
@@ -0,0 +1,77 @@
+Certificate:
```
goloroden, line 1: To be in line with other places where we use certificates, this should rather be `certificate.pem`.

docker/wolkenkit-postgres-ssl/privkey.pem
```diff
@@ -0,0 +1,30 @@
+-----BEGIN ENCRYPTED PRIVATE KEY-----
```
goloroden, line 1: What's the difference between this private key and the other one?

lib/stores/domainEventStore/Postgres/PostgresDomainEventStoreOptions.ts
```diff
@@ -7,6 +8,6 @@ export interface PostgresDomainEventStoreOptions {
   userName: string;
   password: string;
   database: string;
-  encryptConnection?: boolean;
+  encryptConnection?: boolean | ConnectionOptions;
```
goloroden, line 11: Same question as above 😉

lib/stores/lockStore/Postgres/PostgresLockStoreOptions.ts
```diff
@@ -7,6 +8,6 @@ export interface PostgresLockStoreOptions {
   userName: string;
   password: string;
   database: string;
-  encryptConnection?: boolean;
+  encryptConnection?: boolean | ConnectionOptions;
```
goloroden, line 11: Once again 😉

lib/stores/priorityQueueStore/Postgres/PostgresPriorityQueueStoreOptions.ts
```diff
@@ -10,6 +11,6 @@ export interface PostgresPriorityQueueStoreOptions<TItem, TItemIdentifier> {
   userName: string;
   password: string;
   database: string;
-  encryptConnection?: boolean;
+  encryptConnection?: boolean | ConnectionOptions;
```
goloroden, line 14: And another one 😉

docker/wolkenkit-postgres-ssl/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM postgres:13.3-alpine
```
goloroden, line 1: We also need to add this `Dockerfile` to the dependabot configuration.

goloroden (changes requested):
@yeldiRium Sorry that it took me that long to finally review the PR. Overall, it looks good, however, I have left a few comments on minor things that caught my eye.

Could you please have a look at them?

lib/stores/consumerProgressStore/Postgres/PostgresConsumerProgressStoreOptions.ts
yeldiRium: Yes, it is `boolean | ConnectionOptions`. This stems from [@types/pg/index.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cc94a59c9f4244b50b6cc46acad46f75fd10aec7/types/pg/index.d.ts#L27).

lib/common/schemas/getConnectionOptionsSchema.ts
yeldiRium: In principle I agree, but we re-use the `ConnectionOptions` type from the Node.js core module `tls`, and these are the terms that module uses. We could instead define our own options and transform them internally to the options `tls` wants, but I think in the context of a database store configuration it is okay to stick with the technical terms used by the database adapter.

docker/wolkenkit-postgres-ssl/privkey.pem
yeldiRium: I think none and this one is unused anyway. I'll remove it.

lib/common/schemas/getConnectionOptionsSchema.ts
goloroden: I think I'd actually prefer the option to have our own options and transform them, first, because IMHO `privateKey` and `certificate` read more nicely than `key` and `cert`, and second, because this way we decouple our public interface from the technical internals of Node.js.

I know, it's not very probably, but if they change their naming, we would either stick with those names (which still do not read that nicely then), or we would adjust ours as well as a breaking change – in the end, because an internal dependency has a breaking change.

What do you think on this?

yeldiRium: These are valid point and I think you're right that it's a good idea to decouple from node's internals. I'll do that.

goloroden: 😊

goloroden (approved):

PR #2433, opened 2021-09-02
Azure file store

kusigit:
Additional file storage adapter.


strangedev:
Hey @kusigit :wave:
Thank you for your PR! Just letting you know we're aware of it. We'll be in touch regarding the review as soon as feasible.

Have a great day!

yeldiRium:
Hi @kusigit :wave: 

Thanks for opening this PR :) I've had a look at it and it seems fine. I took the liberty of moving the construction logic of the file store instance into the `create` method, so that the structure is a bit more in line with how we create these kinds of instances.

I have only two questions before we can move forward with this:

Firstly, is there a specific reason for using `streamToBuffer`? The only usage I see (https://github.com/thenativeweb/wolkenkit/pull/2433/files#diff-588ee2b4e60364147a119d194ce3df3d4217ffc6b0a020f733b2fe6be61bb3c8R145-R147) converts a `ReadableStream` to a `Buffer` and then converts it back into a `ReadableStream`. Why not just return the stream?

And secondly we need tests to merge this. Is it possible to start an azure file store or an equivalent locally, preferably with docker? If so, we could quite easily integrate that into our test setup for the filestores at https://github.com/thenativeweb/wolkenkit/tree/main/test/integration/stores/fileStore.

Thanks again,
yeldiR

strangedev:
Maybe we can use [Azurite](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite) for testing.

kusigit:
Hi @strangedev 
Perfect, thanks for the hint.

kusigit:
Hi @yeldiRium 

Please create a new repository _wolkenkit-azurite_ in your docker hub and change after the following line:
https://github.com/thenativeweb/wolkenkit/pull/2433/files#diff-649a0c597d8ebb12dccd40ad68ffd6e7bf30c5ba607c3d961a8bae407cf898a7R23

Thanks a lot

Dockerfile
```
FROM mcr.microsoft.com/azure-storage/azurite

# Blob Storage Port
EXPOSE 10000
# Queue Storage Port
EXPOSE 10001
# Table Storage Port
EXPOSE 10002

CMD ["azurite", "-L", "-l", "/data", "--blobHost", "0.0.0.0","--queueHost", "0.0.0.0", "--tableHost", "0.0.0.0"]
```

kusigit:
Hi @yeldiRium 

From my opinion everything in this pull request is fixed. Including tests with Azurite. Please can you have a look again.

Thanks

yeldiRium:
Hi @kusigit,

sorry for the delay. I had another look at it and I'm mostly happy with the current state of things. Using azurite was a good way to go and it seems to work well.

The only thing I'm not a fan of is using the very niche and seemingly unmaintained library `@jorgeferrero/stream-to-buffer` for something that is rather simple. But I also don't just want to copy/rewrite their code into a utility function in the wolkenkit. @goloroden what do you think about this?

Apart from that, this PR can be merged in my opinion. I have already merged the current main branch in preparation.

goloroden:
Closed as discussed via mail.

PR #2651, opened 2022-01-15
fix: repair old REST API calls in templates #1941

### LINKED ISSUE #1941: The client generated by init does not use the current API routes.
### What is this bug about?

When we bootstrap the chat demo application by `npx wolkenkit@4.0.0-internal.108 init`, the client uses old API routes, causing a 404 error.

### What is the expected result?

When we bootstrap the chat demo application, it works as-is.

### What is the actual result?

The chat demo application is not usable, since the client can't fetch from the back-end.

### What steps are needed to reproduce the bug?

- `npx wolkenkit@4.0.0-internal.108 init`
- `npm install`
- `npx wolkenkit dev`
- `cd client && npx http-server`

### What else should we know?

In some internal release, we started differentiating between streaming APIs and APIs that just return a value. Adding `/stream` or `/value` to the paths fixes the issue.

ClemensSahs:
Fixes #1941


goloroden (approved):

goloroden:
@ClemensSahs Thanks for the PR, and I am very sorry that it took us that long to review it. However, the changes look great, and I have now finally approved it.

One thing before I am going to merge it: Would you like to add yourself as a contributor to the `package.json` file? If so, feel free to add to this PR – if not, just let me know and I'll merge without that.

Thanks again for contributing 😊

PR #3018, opened 2022-12-14
Correct handling of async isAuthorized method

kusigit:


goloroden:
Closed as discussed via mail.

